PR #15, opened 2015-05-19
Example still refering to 1.4.2

nellicus:
Example still refering to 1.4.2 and not consistent with parameters reference host => hosts


jsvd:
I took your commit from this pr and included it in #17, closing this one

PR #22, opened 2015-08-28
refactor request logic into own method.

### LINKED ISSUE #21: Out of Memory: Plugin retains too much memory per scan request
By re-using the variable `r` in the `run` method, JRuby does not efficiently clean enough of the data out of memory:

just by refactoring the request to be handled within a more inner and temporary scope of another function, memory usage is reduced radically:

```
def run_next(output_queue, scroll_id)
    r = scroll_request(scroll_id)
    r['hits']['hits'].each do |hit|
      event = LogStash::Event.new(hit['_source'])
      decorate(event)

      if @docinfo
        event[@docinfo_target] ||= {}

        unless event[@docinfo_target].is_a?(Hash)
          @logger.error("Elasticsearch Input: Incompatible Event, incompatible type for the `@metadata` field in the `_source` document, expected a hash got:", :metadata_type => event[@docinfo_target].class)

          raise Exception.new("Elasticsearch input: incompatible event") 
        end

        @docinfo_fields.each do |field|
          event[@docinfo_target][field] = hit[field]
        end
      end
      output_queue << event
    end

    [r['hits']['hits'].any?, r['_scroll_id']]
  end
```

This function is hacky, but it paints a picture in terms of memory performance.

before this, runtime memory usage of a specific config reached `1457 MB`. after this refactor, that same config's memory usage was reduced to a max of `642 MB` on my local computer.

re: https://github.com/logstash-plugins/logstash-input-elasticsearch/blob/master/lib/logstash/inputs/elasticsearch.rb#L152

talevy:
Main reasoning for this refactor is for a lower
memory footprint since response is not assigned to a variable
inside of the #run scope anymore

Fixes #21.

### DIFF
--- a/lib/logstash/inputs/elasticsearch.rb
+++ b/lib/logstash/inputs/elasticsearch.rb
@@ -140,6 +140,32 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
 
     @client = Elasticsearch::Client.new(:hosts => hosts, :transport_options => transport_options)
   end
+
+  private
+  def run_next(output_queue, scroll_id)
+    r = scroll_request(scroll_id)
+    r['hits']['hits'].each do |hit|
+      event = LogStash::Event.new(hit['_source'])
+      decorate(event)
+
+      if @docinfo
+        event[@docinfo_target] ||= {}
+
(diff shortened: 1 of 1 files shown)


suyograo:
LGTM

PR #24, opened 2015-08-31
fix name for non scan configurations

talevy:


talevy:
I am also open to having a check whether `resp` exists. and short-circuiting it to avoid error. what do you think @jsvd?

jsvd:
I like reusing the variable, but unless `run_next`returns an hash with strings as keys then this can break.

I don't see a big performance cost here, but if you consider it big enough then wrap the first scroll response into a symbol-keyed hash and we're done.

talevy:
updated to keep same field name

jsvd:
hum..still, without a `@scan` then search response hash won't have a `"has_hits"` key/value

jsvd:
maybe this should have a test? :) clearly there's some fallible logic here

jsvd:
already included in #27

PR #27, opened 2015-09-01
Fix specs and scan

jsvd:
* refactors specs to use the input spec helper in devutils
* also slightly adapts the scan vs scroll logic to ensure scan works.
* removes the :elasticsearch tag from 2 tests, that can be run offline without elasticsearch

This requires (actually includes) #24

### DIFF
--- a/lib/logstash/inputs/elasticsearch.rb
+++ b/lib/logstash/inputs/elasticsearch.rb
@@ -141,32 +141,6 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
     @client = Elasticsearch::Client.new(:hosts => hosts, :transport_options => transport_options)
   end
 
-  private
-  def run_next(output_queue, scroll_id)
-    r = scroll_request(scroll_id)
-    r['hits']['hits'].each do |hit|
-      event = LogStash::Event.new(hit['_source'])
-      decorate(event)
-
-      if @docinfo
-        event[@docinfo_target] ||= {}
-
-        unless event[@docinfo_target].is_a?(Hash)
-          @logger.error("Elasticsearch Input: Incompatible Event, incompatible type for the `@metadata` field in the `_source` document, expected a hash got:", :metadata_type => event[@docinfo_target].class)
-
-          raise Exception.new("Elasticsearch input: incompatible event")
-        end
-
-        @docinfo_fields.each do |field|
-          event[@docinfo_target][field] = hit[field]
-        end
-      end
-      output_queue << event
-    end
-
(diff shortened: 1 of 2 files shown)


jsvd:
@talevy can you take a look at this? it's now possible to run 2 specs in the suite with just `bundle exec rspec`

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -175,15 +149,48 @@ def run(output_queue)
...
 
   private
+  def run_next(output_queue, scroll_id)
+    r = scroll_request(scroll_id)
+    r['hits']['hits'].each { |hit| push_hit(hit, output_queue) }
+    {'has_hits' => r['hits']['hits'].any?, '_scroll_id' => r['_scroll_id']}
+  end
+
+  private
+  def push_hit(hit, output_queue)
```
talevy: good call

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -175,15 +149,48 @@ def run(output_queue)
 
     # since 'scan' doesn't return data on the search call, do an extra scroll
     if @scan
-      resp = run_next(output_queue, r['_scroll_id'])
+      r = run_next(output_queue, r['_scroll_id'])
+      has_hits = r['has_hits']
+    else # not a scan, process the response
+      r['hits']['hits'].each { |hit| push_hit(hit, output_queue) }
```
talevy: order of `output_queue` and `hit` are swapped

talevy: woops. ignore me! mixed up push_hit and run_next

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -175,15 +149,48 @@ def run(output_queue)
...
-    while resp[:has_hits] do
-      resp = run_next(output_queue, resp[:scroll_id])
+    while has_hits do
+      r = run_next(output_queue, r['_scroll_id'])
+      has_hits = r['has_hits']
     end
   end # def run
 
   private
+  def run_next(output_queue, scroll_id)
```
talevy: I know this is 100% my fault :) but maybe we should rename this method to `process_next_scroll` or something like that. what do you think? seeing run_next in these random branchings confuse me as to what I'm actually calling

talevy:
besides my comments, I like it!

jsvd:
Updated :+1:

talevy:
have you tested against real cluster?

haven't run. but LGTM

jsvd:
I haven't yet, will do tomorrow.

PR #28, opened 2015-09-08
use plugin.stop to return from run

jsvd:
Implements `stop` to return from `run` according to https://github.com/elastic/logstash/issues/3210

Depends on https://github.com/elastic/logstash-devutils/pull/32 and https://github.com/elastic/logstash/pull/3812


talevy:
LGTM

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -157,6 +157,7 @@ def run(output_queue)
     end
 
     while has_hits do
+      return if stop?
```
talevy: why not `whie has_hits && !stop?`?

PR #35, opened 2015-12-07
do not assume event field ref to return in-place mutable structure

colinsurprenant:
- few code cleanups
- reworked field assignment to not assume in-place mutability of Event inner structure

relates to elastic/logstash#4264 and elastic/logstash#4191

### DIFF
--- a/lib/logstash/inputs/elasticsearch.rb
+++ b/lib/logstash/inputs/elasticsearch.rb
@@ -60,7 +60,7 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
   # It might be important to note, with regards to metadata, that if you're
   # ingesting documents with the intent to re-index them (or just update them)
   # that the `action` option in the elasticsearch output want's to know how to
-  # handle those things. It can be dynamically assigned with a field 
+  # handle those things. It can be dynamically assigned with a field
   # added to the metadata.
   #
   # Example
@@ -86,7 +86,7 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
   config :docinfo, :validate => :boolean, :default => false
 
   # Where to move the Elasticsearch document information by default we use the @metadata field.
-  config :docinfo_target, :validate=> :string, :default => "@metadata"
+  config :docinfo_target, :validate=> :string, :default => LogStash::Event::METADATA
 
   # List of document metadata to move to the `docinfo_target` field
   # To learn more about Elasticsearch metadata fields read
@@ -105,7 +105,6 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
   # SSL Certificate Authority file
   config :ca_file, :validate => :path
 
-  public
   def register
     require "elasticsearch"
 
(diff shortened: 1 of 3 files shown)


colinsurprenant:
@guyboertje friendly review ping!

purbon:
please jenkins, test this.

purbon:
please jenkins, test this.

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -160,38 +157,42 @@ def run(output_queue)
...
     decorate(event)
 
     if @docinfo
-      event[@docinfo_target] ||= {}
+      # do not assume event[@docinfo_target] to be in-place updatable. first get it, update it, then at the end set it in the event.
+      docinfo_target = event[@docinfo_target] || {}
 
-      unless event[@docinfo_target].is_a?(Hash)
-        @logger.error("Elasticsearch Input: Incompatible Event, incompatible type for the `@metadata` field in the `_source` document, expected a hash got:", :metadata_type => event[@docinfo_target].class)
+      unless docinfo_target.is_a?(Hash)
```
guyboertje: Given that the only API we require from whatever `docinfo_target` is `[]=` maybe that is a better test.  Also in line 176 we are quite comfortable with making it a new Hash if it is nil consider this:
```ruby
docinfo_target = event[@docinfo_target]
unless docinfo_target.respond_to?("[]=")
  @logger.warn("Elasticsearch Input: Incompatible type for the docinfo_target=#{@docinfo_target} field in the `_source` document, using a hash instead of:", :docinfo_target_type => docinfo_target.class, :event => event)
  docinfo_target = {}
end
```
For me, this means we try our utmost to ingest.
WDYT?

guyboertje:
My comment consideration aside, LGTM

lib/logstash/inputs/elasticsearch.rb
colinsurprenant: hmm... is can only be a `Hash` or an `Array` (or a `String` or numeric). I think its just easier to understand and probably *faster* to use `is_a?` than using `respond_to?`?

guyboertje: True about performance (and btw String responds to `[]=`, what a mess that would make :smile: ).
What about the bit on not raising and forcing a Hash?

colinsurprenant: well, if its not a hash, it is because it is an existing value and the only thing we can do is:
- overwrite and log
- not overwrite/ignore and log
- raise and log

overwriting will destroy an existing value and raising crashes the whole thing. since it currently raises, changing that will change the behaviour of the plugin, but on the other hand, I think this really is an edge case. 

so, I'm on the fense. I'd rather stay on the "safe" side and keep curent behaviour and re-evaluate later in another issue?

colinsurprenant:
Thanks @guyboertje - I'll go ahead and bump version and merge. I'll open a followup issue on the raise discussion.

colinsurprenant:
pushed version 2.0.3

colinsurprenant:
followup issue #36

PR #52, opened 2016-11-24
Update to docs for SSL ca_file option

geekpete:
Make the expected file format and requirement for any chain certificates clearer.

Open to improvements on the wording of this doco text.


talevy:
LGTM

PR #56, opened 2017-02-06
Always use Content-Type application/json

andrewvc:
This is what ES wants/expects


jordansissel:
Is this PR still needed with the fix in upstream elasticsearch-ruby?

PR #57, opened 2017-02-07
Upgrade es client

### LINKED ISSUE #55: Set Content-Type and Accept headers
This is necessary to support https://github.com/elastic/elasticsearch/pull/22691 when it lands. Without fixing this, this input will (likely?) not work against new Elasticsearch releases.

andrewvc:
Sets content-type to application/json

Fixes #55


karmi:
https://rubygems.org/gems/elasticsearch-transport/versions/5.0.3 released which addresses a bug in the Manticore adapter.

PR #60, opened 2017-03-25
Add plugin level required to support sending content-type headers

dedemorton:
Adds compatibility note requested in elastic/logstash#6718

@acchen97 Here's the one for the input plugin for review.


acchen97:
LGTM.

PR #61, opened 2017-04-06
Fix scrolling requests by using json formatted bodies

jsvd:
This makes scrolling incompatible with ES 1.x
NOTE: It's already documented that version 5.x of elasticsearch-ruby gem isn't compatible with ES 1.x, so it should be ok to break scrolling here.

fixes https://github.com/logstash-plugins/logstash-input-elasticsearch/issues/62


andrewvc:
I'll add that this bug only affects 5.3+, ES 5.2 and earlier were not affected.

PR #69, opened 2017-05-23
fix on example code

orhiee:
elk on 5.4.0 gives error if there is a "query => '*' "line


kares:
thanks for the PR - going to close, I think this is defunct by now

Issue #76, opened 2017-08-03
'add_field' and 'tags' can't reference metadata fields


magnusbaeck:
The plugin's push_hit method basically looks like this:

```ruby
def push_hit(hit, output_queue)
    event = LogStash::Event.new(hit['_source'])
    decorate(event)

    # populate metadata fields if docinfo is enabled

    output_queue << event
  end
```

Since we're calling decorate(event) prior to the population of the metadata fields any `add_field` or `tags` options can't reference the metadata fields. Consequently, this configuration won't work:

```plain
input {
  elasticsearch {
    ...
    docinfo => true
    add_field => {
      'some-field-name' => "%{[@metadata][_index]}"
    }
  }
}
```

Postponing the decorate(event) call until after the metadata population should fix this.

PR #78, opened 2017-10-26
enable use of metadata fields in `add_field` decorations

yaauie:
By performing event decorations after the metadata field has been populated
we enable the supported `add_field` decoration to reference metadata fields.

Resolves logstash-plugins/logstash-input-elasticsearch#76


jsvd:
test failure is unrelated to this pr, caused by https://github.com/elastic/logstash/issues/8518

robbavey:
@jsvd @yaauie Should the changelog and patch versions be updated too?

Issue #71, opened 2017-05-31
Support for client certificate authentication


jammann:
I wanted to create a pipeline from one cluster to another with logstash. Both clusters are protected behind reverse proxies, which only speak SSL and require client certificates

The logstash-output-elastic nicely support that (keystore/truststore etc. properties) while the input only provides the ssl boolean config and the ca_certs PEM file. I did not find a way to configure proper client certificate authentication.

Could this feature be added or is there a fundamental issue which prevents that?

- Version: 5.4.0
- Operating System: Linux

PR #80, opened 2017-10-26
TLS client certificate

adrianlzt:
Add options to be able to login to ES using client certificates.
Also an option to ignore checking certificate validity.

### DIFF
--- a/lib/logstash/inputs/elasticsearch.rb
+++ b/lib/logstash/inputs/elasticsearch.rb
@@ -111,8 +111,17 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
   # SSL
   config :ssl, :validate => :boolean, :default => false
 
+  # SSL verify certificate
+  config :ssl_verify, :validate => :boolean, :default => true
+
   # SSL Certificate Authority file in PEM encoded format, must also include any chain certificates as necessary 
   config :ca_file, :validate => :path
+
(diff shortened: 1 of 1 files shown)


adrianlzt:
Example config:
```
input {
  elasticsearch {
    hosts => "elastic:443"
    query => ''
    index => 'test'
    ssl => true
    ssl_verify => false
    ca_file => '/certs/admin-ca'
    client_cert_file => '/certs/admin-cert'
    client_key_file => '/certs/admin-key'
  }
}
```

adrianlzt:
Fixes #71

andrewvc:
Thanks for the patch.

The tricky thing here is adding options consistently across the ES input/output/filter

This is a bit of a weird time as we're trying to centralize into one repo here: https://github.com/logstash-plugins/logstash-integration-elasticsearch/issues/1

Would you mind holding off on this patch till that happens? Otherwise we'll need three simultaneous PRs. We don't want to continue our current pattern of inconsistently adding features.

adrianlzt:
Sure, no problem waiting.

roniez:
This function is highly sought after.

hansvandijk93:
Are there still plans to introduce this feature? 

The feature https://github.com/logstash-plugins/logstash-integration-elasticsearch/issues/1 was mentioned already more than a year ago with no recent updates. So I am curious if there are plans for this in the nearby future.

PR #92, opened 2018-03-15
Improve description of the index

rolfchess:
Directs the user to information on how to reference multiple indices.

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -194,7 +194,7 @@ can be either IP, HOST, IP:port, or HOST:port. The port defaults to
   * Default value is `"logstash-*"`
 
 The index or alias to search. See
-https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-index.html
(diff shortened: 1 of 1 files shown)


rolfchess:
I've signed the  CLA a few hours ago but the check by @karmi in this pull request does not seem to detect it yet. Is there something extra I need to do after signing the CLA?

docs/index.asciidoc
```diff
@@ -193,7 +193,11 @@ can be either IP, HOST, IP:port, or HOST:port. The port defaults to
   * Value type is <<string,string>>
   * Default value is `"logstash-*"`
 
-The index or alias to search.
+The index or alias to search. See
```
jsvd: You can use a named link like so:

```
The index or alias to search. See https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-index.html[Elasticsearch Multi Indices documentation] for more information on how to reference
multiple indices.
```

This way we avoid having a full blown url in the documentation text

rolfchess: I agree, thanks for noticing!

PR #94, opened 2018-07-27
Add managed multi-slice scrolling with `slices` option.

yaauie:
Adds a `slices` option; when given, the input will use the Sliced Scroll api to consume the query via multiple slices in parallel.

I'm not entirely sure the best way to add tests. When implemented as unit tests, I end up having to stub out a series of client calls and responses, which leads me to the point where I'm unsure whether I'm testing the code or my stubs.

### DIFF
--- a/lib/logstash/inputs/elasticsearch.rb
+++ b/lib/logstash/inputs/elasticsearch.rb
@@ -1,6 +1,7 @@
 # encoding: utf-8
 require "logstash/inputs/base"
 require "logstash/namespace"
+require "logstash/json"
 require "base64"
 
 # .Compatibility Note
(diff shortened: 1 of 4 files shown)


lib/logstash/inputs/elasticsearch.rb
```diff
@@ -196,8 +205,25 @@ def stop
...
   def do_run(output_queue)
-    # get first wave of data
-    r = @client.search(@options)
+    # if configured to run a single slice, don't bother spinning up threads
+    return do_run_slice(output_queue) if @slices.nil? || @slices <= 1
+
+    logger.warn("managed slices for query is very large (#{@slices}); consider reducing") if @slices > 8
+
+    @slices.times.map do |slice_id|
+      Thread.new do
```
jsvd: let's name these threads so in the future we'll be able to identify them easily in a thread dump

PR #95, opened 2018-09-04
bump version for 4.3.0 release

yaauie:


jsvd:
https://travis-ci.org/logstash-plugins/logstash-input-elasticsearch/jobs/424562534#L1173 looks like a legit failure

jsvd:
for now removing this from jarvis queue unti we understand the reasons for ci failing

yaauie:
One of the failures is definitely a race condition in the rspec mocks that I had _though_ that I had eliminated, but is apparently still present.

I think the other one is too.

I'll see if I can mitigate. Testing concurrency is hard when RSpec isn't threadsafe.

yaauie:
I'm holding off on merging until I can finish chasing down the race condition. I am confident that it exists only in specs (the validation of partial mocks created with `expect(x).to receive(y)...` is not threadsafe), but am adverse to having flaky specs.

yaauie:
Failures here are due to https://github.com/rspec/rspec-mocks/pull/1236 and will remain intermittent until that fix makes its way to a release. I have not found a way to completely eliminate without solving the underlying bug in rspec, so I will continue to chase that down.

PR #97, opened 2018-09-06
make ssl verification mode configurable

yaauie:
While disabling SSL Certificate Verification eliminates most of the benefits
of using TLS and should _rarely_ be used in production environments, doing so
is occasionally useful in debugging and testing.

This changeset ports the `ssl_certificate_verification` option that is
available in the Elasticsearch Output Plugin, including a similar warning
that is emitted to the plugin's logger.

It also adds specifications how the `ssl` and `ssl_certificate_verification`
directives propagate to the arguments used to create the client.


yaauie:
Same build failure as #95, working to resolve both.

PR #104, opened 2019-05-01
Remove Hash#except from failing spec.

robbavey:
Remove use of Hash#except, which was previously brought in via the
 'i18n' gem, but, as of 7.0 no longer appears to be the case, and is
 causing test failures


robbavey:
Thanks @yaauie - build is green, so I'll go ahead and merge 

(and thank you too @karenzone)

PR #110, opened 2019-10-23
Loosen checks against ES gem and add integration test and docker testing

robbavey:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,19 @@
 ---
-sudo: false
-language: ruby
-cache: bundler
-matrix:
-  include:
-  - rvm: jruby-9.1.13.0
-    env: LOGSTASH_BRANCH=master
-  - rvm: jruby-9.1.13.0
-    env: LOGSTASH_BRANCH=7.0
-  - rvm: jruby-9.1.13.0
-    env: LOGSTASH_BRANCH=6.7
-  - rvm: jruby-9.1.13.0
-    env: LOGSTASH_BRANCH=6.6
-  - rvm: jruby-1.7.27
-    env: LOGSTASH_BRANCH=5.6
-  fast_finish: true
-install: true
-script: ci/build.sh
-jdk: openjdk8
-before_install: gem install bundler -v '< 2'
+sudo: required
+services: docker
+addons:
+  apt:
+    packages:
+    - docker-ce
+env:
+  - INTEGRATION=false ELASTIC_STACK_VERSION=5.x
+  - INTEGRATION=false ELASTIC_STACK_VERSION=6.x
+  - INTEGRATION=false ELASTIC_STACK_VERSION=7.x
+  - INTEGRATION=false ELASTIC_STACK_VERSION=7.x SNAPSHOT=true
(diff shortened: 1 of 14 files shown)


robbavey:
To Reviewer: The build failure against Elastic Stack version 8.x is expected - the Elasticsearch gem currently bundled with Logstash requires a type to be set, while Elasticsearch 8.x does not allow one to be set

.travis.yml
```diff
@@ -1,21 +1,18 @@
...
+    - docker-ce
+env:
+  - INTEGRATION=false ELASTIC_STACK_VERSION=5.x
+  - INTEGRATION=false ELASTIC_STACK_VERSION=6.x
+  - INTEGRATION=false ELASTIC_STACK_VERSION=7.x
+  - INTEGRATION=true ELASTIC_STACK_VERSION=6.x
+  - INTEGRATION=true ELASTIC_STACK_VERSION=7.x
+  - INTEGRATION=true ELASTIC_STACK_VERSION=7.x SNAPSHOT=true
+  - INTEGRATION=true ELASTIC_STACK_VERSION=8.x SNAPSHOT=true
+  - INTEGRATION=true ELASTIC_STACK_VERSION=7.x
```
yaauie: duplicate

```suggestion
```

.travis.yml
```diff
@@ -1,21 +1,18 @@
...
+sudo: required
+services: docker
+addons:
+  apt:
+    packages:
+    - docker-ce
+env:
+  - INTEGRATION=false ELASTIC_STACK_VERSION=5.x
+  - INTEGRATION=false ELASTIC_STACK_VERSION=6.x
+  - INTEGRATION=false ELASTIC_STACK_VERSION=7.x
```
yaauie: missing non-integration snapshots of 7.x and 8.x:
```suggestion
  - INTEGRATION=false ELASTIC_STACK_VERSION=7.x
  - INTEGRATION=false ELASTIC_STACK_VERSION=7.x SNAPSHOT=true
  - INTEGRATION=false ELASTIC_STACK_VERSION=8.x SNAPSHOT=true
```

ci/docker-run.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# This is intended to be run inside the docker container as the command of the docker-compose.
+set -ex
+if [ "$INTEGRATION" == "true" ]; then
```
yaauie: Would it be better to have and use a separate `docker-compose.yml` for integration tests, since non-integration tests cherry-pick a single service to run out of the composition?

spec/es_helper.rb
```diff
@@ -0,0 +1,43 @@
+module ESHelper
+  def self.get_host_port
```
yaauie: from what I can see, this is only ever called from integration tests by way of `ESHelper#get_client`. Is this needed for non-integration? Should it error instead of providing the localhost value that isn't tested?

jsvd:
please refer to the comment in https://github.com/logstash-plugins/logstash-filter-elasticsearch/pull/120#event-2791174338

PR #111, opened 2019-11-19
Switch to using manticore

robbavey:
This commit also adds more integration tests, and pins faraday to `0.15` to ensure continued compatibility with Logstash `5.x`


lib/logstash/inputs/elasticsearch.rb
```diff
@@ -149,6 +150,8 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
   def register
     require "elasticsearch"
     require "rufus/scheduler"
+    require "elasticsearch/transport/transport/http/manticore"
+
```
colinsurprenant:
```suggestion
```

PR #118, opened 2020-03-05
Test: avoid insist usage + load shared example

kares:
for devutils 2.0 compatibility

### DIFF
--- a/spec/inputs/elasticsearch_spec.rb
+++ b/spec/inputs/elasticsearch_spec.rb
@@ -1,5 +1,6 @@
 # encoding: utf-8
 require "logstash/devutils/rspec/spec_helper"
+require "logstash/devutils/rspec/shared_examples"
 require "logstash/inputs/elasticsearch"
 require "elasticsearch"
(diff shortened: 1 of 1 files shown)


kares:
CI failures seem out-of scope. 
**logstash-devutils** wout version specifier simply installs 1.3.6 on 5.x and 2.0.0 on 7.x (both :green_apple:)
all playing nicely for the plugin to do the logstash-core >= version bump later ...

robbavey:
@kares - I think the test failures are related, or at least not fixed by the changes (at least the ones for `7.x` and `8.x`). Running locally, pinning `devutils` to `~>1.3` sees the tests pass.

kares:
Thanks Rob, missed those as I went through CI builds. 
They are very related to the Java pipeline upgrade in devutils 2.0.
Seems there's a race of sorts with the test pipeline setup in specs so we'll need to adapt.

kares:
devutils **2.0.1** is out, let's have a new CI build to verify all is :green_heart:

PR #128, opened 2020-05-21
Add robustness around scrolling

colinsurprenant:
Fixes #122 

- Explicitly call the ES client `clear_scroll` API after each search operation
- Handle scroll_request exceptions

PR #131, opened 2020-05-29
add support for api_key authentication

colinsurprenant:
This is currently based on  #128 which is pending LGTM  - I will rebase once merged.

Related to https://github.com/elastic/logstash/issues/11788

- Add support for [elasticsearch API key](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-create-api-key.html) authentication using the `api_key` option
  - SSL must be enabled
  - Only one authentication method must be specified out of (`user`/`password`, `cloud_auth` and `api_key`)

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -68,6 +68,16 @@ Further documentation describing this syntax can be found
 https://github.com/jmettraux/rufus-scheduler#parsing-cronlines-and-time-strings[here].
 
 
+[id="plugins-{type}s-{plugin}-auth"]
+==== Authentication
+
+Authentication to a secure Elasticsearch cluster is possible using _one_ of the following options:
+
+* <<plugins-{type}s-{plugin}-user>> AND <<plugins-{type}s-{plugin}-password>>
+* <<plugins-{type}s-{plugin}-cloud_auth>>
+* <<plugins-{type}s-{plugin}-api_key>>
+
+
 [id="plugins-{type}s-{plugin}-options"]
 ==== Elasticsearch Input Configuration Options
 
@@ -76,6 +86,7 @@ This plugin supports the following configuration options plus the <<plugins-{typ
 [cols="<,<,<",options="header",]
 |=======================================================================
 |Setting |Input type|Required
+| <<plugins-{type}s-{plugin}-api_key>> |<<password,password>>|No
 | <<plugins-{type}s-{plugin}-ca_file>> |a valid filesystem path|No
 | <<plugins-{type}s-{plugin}-cloud_auth>> |<<password,password>>|No
 | <<plugins-{type}s-{plugin}-cloud_id>> |<<string,string>>|No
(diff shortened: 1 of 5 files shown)


lib/logstash/inputs/elasticsearch.rb
```diff
@@ -351,6 +343,57 @@ def hosts_default?(hosts)
     hosts.nil? || ( hosts.is_a?(Array) && hosts.empty? )
   end
 
+  def validate_authentication
```
yaauie, line 346: same note as in https://github.com/logstash-plugins/logstash-filter-elasticsearch/pull/132 about surprise side-effects from a method with this name

docs/index.asciidoc
```diff
@@ -68,6 +68,12 @@ Further documentation describing this syntax can be found
 https://github.com/jmettraux/rufus-scheduler#parsing-cronlines-and-time-strings[here].
 
 
+[id="plugins-{type}s-{plugin}-auth"]
+==== Authentication
+
+Authentication to a secure Elasticsearch cluster is possible using one of the `user`/`password`, `cloud_auth` or `api_key` options.
```
yaauie, line 74: Same suggestion as in logstash-plugins/logstash-filter-elasticsearch#132

```suggestion
Authentication to a secure Elasticsearch cluster is possible using _one_ of the following options:
* <<plugins-{type}s-{plugin}-user>> AND <<plugins-{type}s-{plugin}-password>>
* <<plugins-{type}s-{plugin}-cloud_auth>>
* <<plugins-{type}s-{plugin}-api_key>>
```

lib/logstash/inputs/elasticsearch.rb
colinsurprenant: As I said, this is also the same in the ES output plugin and I would prefer to keep them aligned. I can change it in both PRs and then submit a patch in the ES output or we defer.

docs/index.asciidoc
colinsurprenant: I'll defer to @karenzone on this!

karenzone: Suggested content is good and will be helpful to users.  Formatting isn't quite right.  For new content to render correctly, you need a return after the colon on line 71.

yaauie: I've updated the suggestion with the newline before the list.

lib/logstash/inputs/elasticsearch.rb
yaauie: +1 to consistency. Ok with resolving in this PR or in a separate effort.

docs/index.asciidoc
karenzone: Fantastic. Thanks!

lib/logstash/inputs/elasticsearch.rb
colinsurprenant: I added it back!

colinsurprenant:
@yaauie @karenzone thanks for the review! - I pushed the discussed changes.

colinsurprenant:
unrelated integration test failures, merging.

colinsurprenant:
v4.7.0 published

Issue #133, opened 2020-07-28
Cannot get new connection from pool when same pipeline read and write to same index


andsel:
It happens that when we have a pipeline composed of elasticsearch input that search all on an index and an elasticsearch output that deletes the documents just loaded from input we get this error:
```
Elasticsearch::Transport::Transport::Error: Cannot get new connection from pool.
  perform_request at /tmp/logstash-7.7.0/vendor/bundle/jruby/2.5.0/gems/elasticsearch-transport-5.0.5/lib/elasticsearch/transport/transport/base.rb:254
  perform_request at /tmp/logstash-7.7.0/vendor/bundle/jruby/2.5.0/gems/elasticsearch-transport-5.0.5/lib/elasticsearch/transport/transport/http/manticore.rb:67
  perform_request at /tmp/logstash-7.7.0/vendor/bundle/jruby/2.5.0/gems/elasticsearch-transport-5.0.5/lib/elasticsearch/transport/client.rb:131
           search at /tmp/logstash-7.7.0/vendor/bundle/jruby/2.5.0/gems/elasticsearch-api-5.0.5/lib/elasticsearch/api/actions/search.rb:183
   search_request at /tmp/logstash-7.7.0/vendor/bundle/jruby/2.5.0/gems/logstash-input-elasticsearch-4.6.0/lib/logstash/inputs/elasticsearch.rb:321
     do_run_slice at /tmp/logstash-7.7.0/vendor/bundle/jruby/2.5.0/gems/logstash-input-elasticsearch-4.6.0/lib/logstash/inputs/elasticsearch.rb:269
           do_run at /tmp/logstash-7.7.0/vendor/bundle/jruby/2.5.0/gems/logstash-input-elasticsearch-4.6.0/lib/logstash/inputs/elasticsearch.rb:257
```

- Version: 4.6.0
- Operating System: Linux
- Steps to Reproduce:
- run an elasticsearch node with docker 
```
docker run -p 9200:9200 -p 9300:9300 -e "discovery.type=single-node" -e "ELASTIC_PASSWORD=changeme" docker.elastic.co/elasticsearch/elasticsearch:7.8.1
```
- fill 10^6 docs with `data_filler_pipeline.conf` later attached
```
bin/logstash -f data_filler_pipeline.conf
```
- use  a pipeline similar to the one provided by the customer
```
bin/logstash -f sut_pipeline.conf
```

Pipeline used to fill data:
```
input {
  generator {
[...]

PR #136, opened 2020-08-20
Fix concurrency bug that avoided to retrieve connection when using multiple slices

andsel:
Patched the RoudRobin policy used by the underlying elasticsearch-transport gem.
In case of multiple slices there are one thread per slice and the library exposed a race condition while not protecting the access to the connections pool.

Fixes: #133


lib/logstash/inputs/patch.rb
```diff
@@ -0,0 +1,44 @@
+if Gem.loaded_specs['elasticsearch-transport'].version >= Gem::Version.new("7.2.0")
```
jsvd, line 1: let's make it clear here that this was fixed already in 7.2.0, but we're using 5.x, and also let's point to the PR where it was solved

jsvd (approved):
LGTM on green CI

PR #140, opened 2020-11-11
Introduce `target` option

yaauie:
Adding a `target` option to the input allows users to specifically target where `_source` of a hit will be placed instead of polluting the top-level of the event.

Combined with the existing `docinfo_fields` and `docinfo_target`, this new setting can be used to ensure that _only_ known fields are populated at the root level, making it easier to define pipelines that are consistently compatible with a schema like ECS.

### DIFF
--- a/lib/logstash/inputs/elasticsearch.rb
+++ b/lib/logstash/inputs/elasticsearch.rb
@@ -175,6 +175,9 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
   # exactly once.
   config :schedule, :validate => :string
 
+  # If set, the _source of each hit will be added nested under the target instead of at the top-level
+  config :target, :validate => :field_reference
+
   def register
     require "elasticsearch"
     require "rufus/scheduler"
@@ -298,7 +301,12 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
   end
 
   def push_hit(hit, output_queue)
-    event = LogStash::Event.new(hit['_source'])
+    if @target.nil?
+      event = LogStash::Event.new(hit['_source'])
+    else
+      event = LogStash::Event.new
+      event.set(@target, hit['_source'])
+    end
 
     if @docinfo
       # do not assume event[@docinfo_target] to be in-place updatable. first get it, update it, then at the end set it in the event.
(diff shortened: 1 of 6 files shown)


lib/logstash/inputs/elasticsearch.rb
```diff
@@ -475,4 +483,31 @@ def validate_value(value, validator)
     end
   end
   extend(PositiveWholeNumberValidator)
+
+  module FieldReferenceValidator
```
yaauie, line 487: I think this needs to be extracted into a separate support mixin, since it will be helpful to use in other plugins implementing `target`.

It would be great if we could provide this in Logstash core, but the way the config mixin works currently is non-extensible, and anything we add to it cannot be consumed by plugins without preventing those plugins from working on older versions of Logstash.

kares (approved):
:+1: simple but so valuable

haven't realized we could be validating field references, bonus points there!

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -298,7 +301,12 @@ def process_next_scroll(output_queue, scroll_id)
   end
 
   def push_hit(hit, output_queue)
-    event = LogStash::Event.new(hit['_source'])
+    if @target.nil?
+      event = LogStash::Event.new(hit['_source'])
+    else
+      event = LogStash::Event.new
```
kares, line 307: I am assuming we deliberately do not want to treat the `'@timestamp': ...` field specially?
`LogStash::Event.new('@timestamp' => hit['_source']['@timestamp'])`

yaauie: I think this needs to be an either/or situation. Either we treat the `_source` _as_ the event (including magical fields) or we target the `_source` _to_ a place on the event (and the user then uses the `date` filter to pull the `@timestamp` over). If we attempt to mix, we will invariably get it wrong for some uses.

kares: OK, I am fine either way. Just more 'natural' for me to re-target regular fields (but not `@timestamp`).
Reading the same document and having different timestamps is surprising behavior despite the `target => ...` setting esp. since I can not have a nil `@timestamp` on the new event. Likely very subjective on my end ...

yaauie: resolving this by leaving as-is. `target` is about creating an event _with_ the `_source` of a hit, and without `target` is the path to create an event _from_ the `_source`.

PR #143, opened 2020-12-15
Doc: Replace cross doc links with attributes

karenzone:
Standardize links to other Elastic docs to use shared attributes
Bump to v.4.9.1


andsel (approved):
LGTM

PR #148, opened 2021-07-09
Switch ES ruby client from >= 5.0.3 to >= 7.13.0

andsel:
Update the ES client to version `7.13.0`


andsel:
#157 requires a more recent version of ES client library

PR #149, opened 2021-07-19
Feat: ecs_compatibility + event_factory support

kares:
Review for ECS support.

The plugin already supports the `target => ...` option, ~~however it's questionable whether we need to warn on `target` not being set in ECS mode (basically whether we want to track ECS mode) since there's a decent chance the source data in ES might already be ECS compliant. Warning due unset target might be confusing.~~

NOTE: the unused JSON codec default was removed since it causes a warning due the unset `target` on the codec.

Also resolves https://github.com/logstash-plugins/logstash-input-elasticsearch/issues/151 (for good)


lib/logstash/inputs/elasticsearch.rb
```diff
@@ -126,7 +133,7 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
   config :docinfo, :validate => :boolean, :default => false
 
   # Where to move the Elasticsearch document information. By default we use the @metadata field.
-  config :docinfo_target, :validate=> :string, :default => LogStash::Event::METADATA
+  config :docinfo_target, :validate=> :field_reference, :default => LogStash::Event::METADATA
```
kares, line 136: there's an opportunity to change the default here in ECS mode to `[@metadata][input][elasticsearch]` 
so we would end up with: `[@metadata][input][elasticsearch][_index]` etc instead of just `[@metadata][_index]`

yaauie: I like it for consistency with the other plugin metadata. 👍🏼

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -63,12 +66,16 @@
...
+  include LogStash::PluginMixins::ECSCompatibilitySupport(:disabled, :v1, :v8 => :v1)
+  include LogStash::PluginMixins::ECSCompatibilitySupport::TargetCheck
+
+  include LogStash::PluginMixins::EventSupport::EventFactoryAdapter
+
   extend LogStash::PluginMixins::ValidatorSupport::FieldReferenceValidationAdapter
 
   config_name "elasticsearch"
 
-  default :codec, "json"
```
yaauie, line 70: Side note: I think we need a way to specify a noop codec. Because the base classes default to `codec => plain`, we instantiate a codec that we will never use during setup of the plugin.

yaauie (commented):
Code LGTM. I like your suggestion to change the default metadata target to align with other plugins. It is more verbose, but alignment is key.

kares:
Thanks, `docinfo_target`'s `@metadata` default now uses our plugin convention in ECS mode.

there's a new `undefined method options for #<Elasticsearch::Transport::Client:0x20797dcc>` in the CI snapshot targets which seems to be related to the elasticsearch client update to **7.14.0** ... will investigate separately.

spec/inputs/elasticsearch_spec.rb
```diff
@@ -756,4 +773,9 @@ def synchronize_method!(object, method_name)
     end
 
   end
+
+  # @note the hosts method got removed and @hosts i-var was changed to @seeds in elasticsearch 7.14.0
```
yaauie, line 777: Looking at the Elasticsearch client libraries, I'm seeing that [7.14.0's Transport class still has `hosts`](https://github.com/elastic/elasticsearch-ruby/blob/v7.14.0/elasticsearch-transport/lib/elasticsearch/transport/transport/base.rb#L53) (and exposes it as an attr reader -- so we shouldn't need the ivar reach-in?). What am I missing?

I _really_ would like to avoid reaching into ivars of libraries we don't own, even in specs.

kares: :+1: it's quite weird esp. since after https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/152 things seemed :green_circle:  
but than merging with master here and :red_circle:  again ...  :disappointed: will have to revisit this once again

kares: probably the object we get has changed in between elasticsearch 5.x -> 7.14.0 
used to return the Transport itself but now it's a [`Transport::Client`](https://github.com/elastic/elasticsearch-ruby/blob/v7.14.0/elasticsearch-transport/lib/elasticsearch/transport/client.rb) (with an extra [`.transport`](https://github.com/elastic/elasticsearch-ruby/blob/v7.14.0/elasticsearch-transport/lib/elasticsearch/transport/client.rb#L68) to get the same as in 5.x)

lib/logstash/inputs/elasticsearch.rb
kares: :+1: here the reason for removal is mostly due json codec warning in ECS mode on an unset `target`

spec/inputs/elasticsearch_spec.rb
kares: there's a work-around for this now that should be cleaner in terms of not reading i-vars.
btw. LS used elasticsearch 5.0.5 up till 7.13.4 (LS 7.14.0 upgraded to 7.5.0).

PR #150, opened 2021-07-28
ECS refinements

yaauie:
- uses shared `TargetCheck` to guide toward usage of the target parameter
 - migrates to event factory, dropping deprecated usage of `LogStash::Event.new`


yaauie:
🤦🏼 duplicate of #149

PR #153, opened 2021-08-11
Fix: client gem 7.x compatibility

kares:
These changes fix the [regression in LS 7.14.0](https://github.com/elastic/logstash/issues/13122) where due the elasticsearch client update (from 5.0.5 to 7.5.0) the `Authorization` header isn't passed (due a [bug](https://github.com/elastic/elasticsearch-ruby/issues/1428) in client versions 7.2.0 - 7.14.0). This leads to the plugin not being able to leverage `user`/`password` credentials set by the user to connect to ES.

The fix involves a patch to work-around the elasticsearch client issue (same one as https://github.com/logstash-plugins/logstash-filter-elasticsearch/pull/148).

The verbosity of changes is due refactoring (integration) testing which turned out to be slightly challenging - but we know have a `SECURE_INTEGRATION` targets in the CI which have proper certificates/ and does not allow connection without authentication.

~~Docker was adjusted to be able to run ES secured with $ELASTIC_PASSWORD for the *elastic* user (for integration tests).~~


andsel (approved):
LGTM,
tested the gem locally with `7.13.4` and `7.14.0` and works great!

PR #156, opened 2021-08-18
Fixed SSL handshake hang indefinitely with proxy setup

kaisecheng:
the fix in upstream has merged and [released](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1032#issuecomment-900782420) in 0.7.1 gem
Hence, update manticore

Fix: https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/1033
Related: https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1032


andsel (approved):
LGTM

PR #157, opened 2021-08-18
Verify connection to Elasticsearch 

andsel:
Switched elasticserach client to 7.14 and added a test connection method during registration


lib/logstash/inputs/elasticsearch.rb
```diff
@@ -460,6 +462,14 @@ def parse_user_password_from_cloud_auth(cloud_auth)
   # @private used by unit specs
   attr_reader :client
 
+  def test_connection!
+    begin
+      @client.ping
+    rescue Elasticsearch::UnsupportedProductError
+      raise LogStash::ConfigurationError, "Not a valid Elasticsearch"
+    end
+  end
```
jsvd, line 471:
```suggestion
  def test_connection!
    @client.ping
  rescue Elasticsearch::UnsupportedProductError
    raise LogStash::ConfigurationError, "Not a valid Elasticsearch"
  end
```

jsvd (commented):
The code change LGTM but we should have a better error message before proceeding with merging.

jsvd (commented):
version 6.x is still supported even in the elasticsearch-ruby client https://github.com/elastic/elasticsearch-ruby/commit/bd796ea22c40987f7e07e83978436e465daf6e6b#diff-f551818499dca113dde958c371a216b7159a3fc33706e82cf25ac88b10b1caf8R68-R70, what is the need to drop testing here?

andsel:
> version 6.x is still supported even in the elasticsearch-ruby client

@jsvd the problem could be seen on the [CI failure](https://app.travis-ci.com/github/logstash-plugins/logstash-input-elasticsearch/jobs/541111792#L285)  and is a conflict between the ES client library requested by the Logstash and the one requested by this plugin. In LS versions <= LS 7.14 this plugin can't be installed because of the conflicting ES client versions.

jsvd:
I see so the problem is that we don't have the ability in the CI scripts to override the ELASTIC_STACK version, which would allow us to test the LS 7.x against a 6.x ES.
This override capability should be limited to the elasticsearch plugins, that said, can you open an issue in logstash-plugins/.ci ?

andsel:
Created https://github.com/logstash-plugins/.ci/issues/25

jsvd (approved):
LGTM

chrishiste:
Hello @andsel and @jsvd. First of all, thanks for your good work on this plugin. 

If I'm understanding correctly this PR, the following line forces to use Elasticsearch 7.14:
CleanShot 2021-12-14 at 12 02 54@2x

I'm still using Elasticsearch 7.10. But I would like to upgrade Logstash to the latest version (7.16.1) in order to protect myself from the recent security issue affecting Log4j2 (CVE-2021-44228). This PR's code has been adopted in the 7.16.0 version of Logstash.

Could you please explain the reason support of older versions was dropped? Is this due to a technical limitation? 

Thanks in advance for your help on this.

andsel:
@chrishiste you are safe to use `elasticsearch` client `7.14` to connect to `7.10`.

The Elasticsearch clients can connect to same major versions of Elasticsearch.

The technical reason is that modern Logstash versions has a [dependency on Elasticsearch 7](https://github.com/elastic/logstash/commit/ef9b0d2db5f1a30cd0cef7920991b3b309323019#diff-db8735f3fd64e29befb30510d77ab93ae863feba2a3349f048dfc50c57df73ec) so we took the opportunity to aligns also the plugins.

chrishiste:
@andsel I see, then something else must be causing my error. Here's what I get `ConfigurationError: Could not connect to a compatible version of Elasticsearch`.  I'm using 7.10 OSS is that considered incompatible?

PR #162, opened 2021-11-29
Test: against ES with configured TLSv1.3

kares:
This PR establishes confidence that TLSv1.3 is working and supported on recent LS (JVM) versions.

An integration target is added where ES 7.x is set to only accept TLS 1.3 protocol and gets tested against LS.


kares:
started bumping into https://github.com/logstash-plugins/logstash-input-elasticsearch/issues/165

kares:
:red_circle: CI bumping into 
- https://github.com/elastic/logstash/issues/13890
- https://github.com/elastic/logstash/issues/13572
- https://github.com/logstash-plugins/logstash-input-elasticsearch/issues/165

.ci/Dockerfile.elasticsearch
```diff
@@ -15,6 +16,9 @@ RUN if [ "$SECURE_INTEGRATION" = "true" ] ; then echo "xpack.security.http.ssl.k
 RUN if [ "$SECURE_INTEGRATION" = "true" ] ; then echo "xpack.security.http.ssl.certificate: $es_path/config/test_certs/es.crt" >> $es_yml; fi
 RUN if [ "$SECURE_INTEGRATION" = "true" ] ; then echo "xpack.security.http.ssl.certificate_authorities: [ '$es_path/config/test_certs/ca.crt' ]" >> $es_yml; fi
 RUN if [ "$SECURE_INTEGRATION" = "true" ] ; then echo "xpack.security.http.ssl.verification_mode: certificate" >> $es_yml; fi
+RUN if [ "$SECURE_INTEGRATION" = "true" ] && [ ! -z "$ES_SSL_SUPPORTED_PROTOCOLS" ] ; then echo "xpack.security.http.ssl.supported_protocols: ${ES_SSL_SUPPORTED_PROTOCOLS}" >> $es_yml; fi
```
kaisecheng, line 19:
```suggestion
RUN if [ "$SECURE_INTEGRATION" = "true" ] && [ -n "$ES_SSL_SUPPORTED_PROTOCOLS" ] ; then echo "xpack.security.http.ssl.supported_protocols: ${ES_SSL_SUPPORTED_PROTOCOLS}" >> $es_yml; fi
```

kaisecheng (commented):
Overall looks good. Most of the updates are from filter-elasticsearch.
The following red CI seems different from the known issues
INTEGRATION=false ELASTIC_STACK_VERSION=7.x
INTEGRATION=false ELASTIC_STACK_VERSION=8.x

kares:
> The following red CI seems different from the known issues
`INTEGRATION=false ELASTIC_STACK_VERSION=7.x`
`INTEGRATION=false ELASTIC_STACK_VERSION=8.x`

Right, there's more than just the interruptible failure mentioned in https://github.com/elastic/logstash/issues/13572
There seems to be a unit test leak but we do not have any production nor unit test related changes here. 
Does not seem to be related to the changes here and I am also not reproducing the leak locally ...

kares:
tried handling the extra spec failures in https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/169

kaisecheng (approved):
LGTM. Red CI is unrelated to this change. All failed CI tests are traced in issues.

Issue #167, opened 2022-02-09
custom user agent no longer set since LS 7.16


kares:
This is a regression with the Elasticsearch(-transport) gem.

Versions 7.14 / 7.15 work as expected but 7.16 and 7.17 started failing and keep using the default header.
Upstream issue: https://github.com/elastic/elasticsearch-ruby/issues/1684

The issue reproduces in CI when running specs:
```
Failures:

  1) LogStash::Inputs::Elasticsearch client 'user-agent' header used by plugin client should sent the expect user-agent
     Failure/Error: expect(request.header['user-agent'][0]).to match(/logstash\/\d*\.\d*\.\d* \(OS=.*; JVM=.*\) logstash-input-elasticsearch\/\d*\.\d*\.\d*/)
     
       expected "elasticsearch-ruby/7.17.0 (RUBY_VERSION: 9.2.20.0; linux x86_64; Manticore 0.8.0)" to match /logstash\/\d*\.\d*\.\d* \(OS=.*; JVM=.*\) logstash-input-elasticsearch\/\d*\.\d*\.\d*/
       Diff:
       @@ -1 +1 @@
       -/logstash\/\d*\.\d*\.\d* \(OS=.*; JVM=.*\) logstash-input-elasticsearch\/\d*\.\d*\.\d*/
       +"elasticsearch-ruby/7.17.0 (RUBY_VERSION: 9.2.20.0; linux x86_64; Manticore 0.8.0)"
       
     # ./spec/inputs/elasticsearch_spec.rb:874:in `block in <main>'
     # /opt/local/rvm/gems/jruby-9.2.20.0@ls/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `block in <main>'
     # /opt/local/rvm/gems/jruby-9.2.20.0@ls/gems/logstash-devutils-2.3.0-java/lib/logstash/devutils/rspec/spec_helper.rb:61:in `block in <main>'

Finished in 2.3 seconds (files took 6.95 seconds to load)
2 examples, 1 failure

Failed examples:

rspec ./spec/inputs/elasticsearch_spec.rb:865 # LogStash::Inputs::Elasticsearch client 'user-agent' header used by plugin client should sent the expect user-agent
```
---

The issue also affects ES filter: https://github.com/logstash-plugins/logstash-filter-elasticsearch/issues/157

andsel:
[...]

Issue #165, opened 2022-02-03
Test failing with JNR error: 


kares:
Tests failing after the Docker base image switch (from CentOS to Ubuntu).
7.17 as well as all 8.x images are hitting a potential JRuby bug, sample output:
```
logstash_1_946bade3a63a | Feb 03, 2022 10:51:54 AM jnr.netdb.NativeProtocolsDB load
logstash_1_946bade3a63a | WARNING: Failed to load native protocols db
logstash_1_946bade3a63a | java.lang.RuntimeException: getprotobyname_r failed
logstash_1_946bade3a63a | 	at org.jruby.dist/jnr.netdb.NativeProtocolsDB$LinuxNativeProtocolsDB.getProtocolByName(NativeProtocolsDB.java:180)
logstash_1_946bade3a63a | 	at org.jruby.dist/jnr.netdb.NativeProtocolsDB.load(NativeProtocolsDB.java:80)
logstash_1_946bade3a63a | 	at org.jruby.dist/jnr.netdb.NativeProtocolsDB.access$000(NativeProtocolsDB.java:40)
logstash_1_946bade3a63a | 	at org.jruby.dist/jnr.netdb.NativeProtocolsDB$SingletonHolder.<clinit>(NativeProtocolsDB.java:47)
logstash_1_946bade3a63a | 	at org.jruby.dist/jnr.netdb.NativeProtocolsDB.getInstance(NativeProtocolsDB.java:43)
logstash_1_946bade3a63a | 	at org.jruby.dist/jnr.netdb.Protocol$ProtocolDBSingletonHolder.load(Protocol.java:107)
logstash_1_946bade3a63a | 	at org.jruby.dist/jnr.netdb.Protocol$ProtocolDBSingletonHolder.<clinit>(Protocol.java:103)
logstash_1_946bade3a63a | 	at org.jruby.dist/jnr.netdb.Protocol.getProtocolDB(Protocol.java:96)
logstash_1_946bade3a63a | 	at org.jruby.dist/jnr.netdb.Protocol.getProtocolByNumber(Protocol.java:59)
logstash_1_946bade3a63a | 	at org.jruby.dist/org.jruby.ext.socket.Addrinfo.<init>(Addrinfo.java:806)
logstash_1_946bade3a63a | 	at org.jruby.dist/org.jruby.ext.socket.SocketUtils$2.addrinfo(SocketUtils.java:255)
[...]

PR #168, opened 2022-02-17
Fix: hosts => "es_host:port" regression

kares:
Same as: https://github.com/logstash-plugins/logstash-filter-elasticsearch/pull/156

> The change here makes sure "es_host:port" is properly split into a `{ host: es_host, port: port }` pair.
Elasticsearch gem not seeing a port: key will add the port: 9200 default and Manticore 0.8 appends the port part while building the URL.

The fix is demonstrated by the `SECURE_INTEGRATION=true` targets : 
- :red_circle: before (the updated `setup_hosts`): https://app.travis-ci.com/github/logstash-plugins/logstash-input-elasticsearch/builds/246624077
- :green_circle: after (updated `setup_hosts`): https://app.travis-ci.com/github/logstash-plugins/logstash-input-elasticsearch/builds/246624839

Also added some explicit CI targets against Manticore 0.7.1 which has been dominant through 7.x line.

---

Unfortunately specs are :red_circle: on CI due various issues:
- https://github.com/logstash-plugins/logstash-input-elasticsearch/issues/167
- https://github.com/logstash-plugins/logstash-input-elasticsearch/issues/165
- https://github.com/elastic/logstash/issues/13572


jsvd (approved):
LGTM

PR #169, opened 2022-03-16
Test: refactor potential stub leak from example

kares:
now there's only the interruptible spec failure (https://github.com/elastic/logstash/issues/13572)
https://app.travis-ci.com/github/logstash-plugins/logstash-input-elasticsearch/jobs/563508142


kaisecheng (approved):
Thanks for fixing the test 🍓

PR #175, opened 2022-03-29
Test: less double leaking failures (on CI)

kares:
Refactoring of test code, avoiding the "old" `input(config) do |pipeline, queue| ...` format.

motivated mostly due (still - https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/169) running into:
```
logstash_1_c3816da54f91 |      Failure/Error: @client.search(options)
logstash_1_c3816da54f91 |        #<Double "elasticsearch-client"> was originally created in one example but has leaked into another example and can no longer be used. rspec-mocks' doubles are designed to only last for one example, and you need to create a new one in each example you wish to use it for.
logstash_1_c3816da54f91 |      # ./lib/logstash/inputs/elasticsearch.rb:361:in `search_request'
logstash_1_c3816da54f91 |      # ./lib/logstash/inputs/elasticsearch.rb:290:in `do_run_slice'
logstash_1_c3816da54f91 |      # ./lib/logstash/inputs/elasticsearch.rb:269:in `do_run'
logstash_1_c3816da54f91 |      # ./lib/logstash/inputs/elasticsearch.rb:252:in `block in run'
logstash_1_c3816da54f91 | 
logstash_1_c3816da54f91 | Finished in 1 minute 4.32 seconds (files took 10.89 seconds to load)
logstash_1_c3816da54f91 | 69 examples, 2 failures

logstash_1_c3816da54f91 | 
logstash_1_c3816da54f91 | Failed examples:
logstash_1_c3816da54f91 | 
logstash_1_c3816da54f91 | rspec ./spec/inputs/elasticsearch_spec.rb:49 # LogStash::Inputs::Elasticsearch behaves like an interruptible input plugin #stop returns from run

logstash_1_c3816da54f91 | rspec './spec/inputs/elasticsearch_spec.rb[1:3:2:1]' # LogStash::Inputs::Elasticsearch `ecs_compatibility => disabled` when a target is set creates the event using the target
```

### DIFF
--- a/spec/inputs/elasticsearch_spec.rb
+++ b/spec/inputs/elasticsearch_spec.rb
@@ -78,14 +78,10 @@ describe LogStash::Inputs::Elasticsearch, :ecs_compatibility_support do
     end
 
     let(:config) do
-      %q[
-        input {
-          elasticsearch {
-            hosts => ["localhost"]
-            query => '{ "query": { "match": { "city_name": "Okinawa" } }, "fields": ["message"] }'
-          }
-        }
-      ]
+      {
+          'hosts' => ["localhost"],
+          'query' => '{ "query": { "match": { "city_name": "Okinawa" } }, "fields": ["message"] }'
+      }
     end
 
     let(:mock_response) do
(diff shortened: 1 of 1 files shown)


spec/inputs/elasticsearch_spec.rb
```diff
@@ -120,40 +123,35 @@
     end
 
     before(:each) do
-      client = Elasticsearch::Client.new
-      expect(Elasticsearch::Client).to receive(:new).with(any_args).and_return(client)
+      expect(Elasticsearch::Client).to receive(:new).with(any_args).and_return client = Elasticsearch::Client.new
```
andsel, line 126: I think the version in 2 lines is more readable, unless for some reason this line is mandatory to fix the problem

andsel (approved):
LGTM

The refactoring rationalize the config variable used in test, exposing only the deltas between the base and the test specific requirement.
It's not clear to me why it reduces the test doubles leaks.

Left a comment, maybe stylistic.

spec/inputs/elasticsearch_spec.rb
kares:
```suggestion
      client = Elasticsearch::Client.new 
      expect(Elasticsearch::Client).to receive(:new).with(any_args).and_return(client)
```

kares:
>  It's not clear to me why it reduces the test doubles leaks.

we won't be starting plugin async-ly in separate pipeline thread (`input(string) ...` from devutils), 
that should improve the predictability of the behavior - most double fails relate to scheduler but they're also seemingly reported from non-scheduler specs :hankey:

PR #177, opened 2022-05-10
Refactor: switch to using scheduler mixin

kares:
#### part of https://github.com/elastic/logstash/issues/13773

see https://github.com/logstash-plugins/logstash-mixin-scheduler/pull/1


lib/logstash/inputs/elasticsearch.rb
```diff
@@ -274,9 +269,10 @@ def do_run(output_queue)
 
     logger.warn("managed slices for query is very large (#{@slices}); consider reducing") if @slices > 8
 
+    pipeline_id = (respond_to?(:execution_context) && execution_context&.pipeline_id) || 'main'
     @slices.times.map do |slice_id|
       Thread.new do
-        LogStash::Util::set_thread_name("#{@id}_slice_#{slice_id}")
+        LogStash::Util::set_thread_name("[#{pipeline_id}]|input|elasticsearch|slice_#{slice_id}")
```
kares, line 275: minor unrelated update - same thread naming strategy as in the scheduler

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -274,9 +269,10 @@ def do_run(output_queue)
 
     logger.warn("managed slices for query is very large (#{@slices}); consider reducing") if @slices > 8
 
+    pipeline_id = (respond_to?(:execution_context) && execution_context&.pipeline_id) || 'main'
```
yaauie, line 272: `LogStash::Plugin#execution_context` has been in place [since 6.0.0](https://github.com/elastic/logstash/commit/0d1011980efc2fdae9b8e8a261b8d04f6ff537de), and it has supported `LogStash::ExecutionContext#pipeline_id` since its introduction. 

In 7.9, we began returning a common [null implementation](https://github.com/elastic/logstash/pull/12305/files) of `ExecutionContext` instead of nil when it wasn't set for the plugin instance.

```suggestion
    pipeline_id = execution_context&.pipeline_id || 'main'
```

yaauie (approved):
Minor nitpick about being overly defensive when building the `pipeline_id` for the thread name, but otherwise LGTM.

PR #178, opened 2022-05-19
ssl: add support for `ca_trusted_fingerprint` on Logstash 8.3+

yaauie:
NOTE:
 - depends on support adapter introduced in https://github.com/logstash-plugins/logstash-mixin-ca_trusted_fingerprint_support/pull/1 to run on any Logstash, regardless of plugin config using the new `ca_trusted_fingerprint` option
 - Use of the `ca_trusted_fingerprint` option works on Logstash 8.3+ after https://github.com/elastic/logstash/pull/14120


mashhurs (approved):
Thanks @yaauie to help on env setups.
I have configured on my local and tested with interactive console. Since it is little complex steps, I have documented how this change can be verified. [DOC LINK](https://docs.google.com/document/d/1wkYkqUW121SbupCIH1ko5vf7s1qsZsC4QvNL7yFwYOE)

PR #179, opened 2022-08-01
add retry for query

kaisecheng:
Prior to this change, throwing exceptions in [search_request() ](https://github.com/logstash-plugins/logstash-input-elasticsearch/blob/657c3ff1b3b6326106b82ce8ca8decb618fa59cd/lib/logstash/inputs/elasticsearch.rb#L361)crashes Logstash as it is [unable to be caught](https://github.com/elastic/logstash/issues/11603) in the [inputworker()](https://github.com/elastic/logstash/blob/v8.3.3/logstash-core/lib/logstash/java_pipeline.rb#L405). Hence, this commit adds retry for error handling for query. 

Add `tries` option for the number of runs 
- should be >= 1
- 2 means having 1 execution and 1 retry

Use Stud::Try for retry in Thread. After retries fail, log an error msg.

### DIFF
--- a/lib/logstash/inputs/elasticsearch.rb
+++ b/lib/logstash/inputs/elasticsearch.rb
@@ -10,6 +10,7 @@ require 'logstash/plugin_mixins/ecs_compatibility_support/target_check'
 require 'logstash/plugin_mixins/ca_trusted_fingerprint_support'
 require "logstash/plugin_mixins/scheduler"
 require "base64"
+require 'logstash/helpers/loggable_try'
 
 require "elasticsearch"
 require "elasticsearch/transport/transport/http/manticore"
@@ -98,6 +99,9 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
 
   # This allows you to set the maximum number of hits returned per scroll.
   config :size, :validate => :number, :default => 1000
+
+  # The number of try for the schedule job
+  config :tries, :validate => :number, :default => 1
 
   # This parameter controls the keepalive time in seconds of the scrolling
   # request and initiates the scrolling process. The timeout applies per
@@ -221,6 +225,8 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
       @slices < 1 && fail(LogStash::ConfigurationError, "Elasticsearch Input Plugin's `slices` option must be greater than zero, got `#{@slices}`")
     end
 
+    @tries < 1 && fail(LogStash::ConfigurationError, "Elasticsearch Input Plugin's `tries` option must be greater than zero, got `#{@tries}`")
+
     validate_authentication
     fill_user_password_from_cloud_auth
     fill_hosts_from_cloud_id
@@ -262,10 +268,12 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
   end
 
   private
-
+  JOB_NAME = "run query"
   def do_run(output_queue)
     # if configured to run a single slice, don't bother spinning up threads
-    return do_run_slice(output_queue) if @slices.nil? || @slices <= 1
+    return retryable(JOB_NAME) do
+      do_run_slice(output_queue)
+    end if @slices.nil? || @slices <= 1
 
     logger.warn("managed slices for query is very large (#{@slices}); consider reducing") if @slices > 8
 
@@ -273,9 +281,24 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
     @slices.times.map do |slice_id|
       Thread.new do
         LogStash::Util::set_thread_name("[#{pipeline_id}]|input|elasticsearch|slice_#{slice_id}")
-        do_run_slice(output_queue, slice_id)
+        retryable(JOB_NAME) do
+          do_run_slice(output_queue, slice_id)
+        end
       end
     end.map(&:join)
+  end
+
+  def retryable(job_name, &block)
+    begin
+      stud_try = ::LogStash::Helpers::LoggableTry.new(logger, job_name)
+      stud_try.try(@tries.times) {
+        block.call
+      }
+    rescue => e
+      error_details = {:message => e.message, :cause => e.cause}
+      error_details[:backtrace] = e.backtrace if logger.debug?
+      logger.error("Tried #{@tries} times #{job_name} unsuccessfully", error_details)
+    end
   end
 
   def do_run_slice(output_queue, slice_id=nil)
--- a/lib/logstash/helpers/loggable_try.rb
+++ b/lib/logstash/helpers/loggable_try.rb
@@ -1,0 +1,18 @@
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+
+require 'stud/try'
+
+module LogStash module Helpers
+  class LoggableTry < Stud::Try
+    def initialize(logger, name)
+      @logger = logger
+      @name = name
+    end
+
+    def log_failure(exception, fail_count, message)
+      @logger.warn("Attempt to #{@name} but failed. #{message}", fail_count: fail_count, exception: exception.message)
+    end
+  end
+end end


kaisecheng:
Red CI is unrelated to the change https://app.travis-ci.com/github/logstash-plugins/logstash-input-elasticsearch/builds

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -262,22 +268,39 @@ def run(output_queue)
...
+          do_run_slice(output_queue, slice_id)
+        end
       end
     end.map(&:join)
   end
 
+  def retryable(job_name, &block)
+    begin
+      stud_try = ::LogStash::Helpers::LoggableTry.new(logger, job_name)
+      stud_try.try(@tries.times) {
```
mashhurs, line 294: Thanks @kaisecheng for taking care this. This will be long lasting task and will give benefit alot!
One quick note at the glance: As I remember, `LogStash::Helpers::LoggableTry` tries on `StandardError` types. Do we need to specify timeout exceptions to make sure it retries only expected cases?
Eg.: `::LogStash::Outputs::ElasticSearch::HttpClient::Pool::HostUnreachableError`

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -262,22 +268,39 @@ def run(output_queue)
   end
 
   private
-
+  JOB_NAME = "run query"
   def do_run(output_queue)
     # if configured to run a single slice, don't bother spinning up threads
-    return do_run_slice(output_queue) if @slices.nil? || @slices <= 1
+    return retryable(JOB_NAME) do
```
andsel, line 274: I think this, with the addition of `retryable` decorator, has switched to a not trivial block and  hit a [style rule](https://github.com/rubocop/ruby-style-guide#multi-line-if-modifiers).
To me is more obscure  than:
```ruby
if @slices.nil? || @slices <= 1
  return retryable(JOB_NAME) do
    do_run_slice(output_queue)
  end
end
```

andsel (changes requested):
Nice feature, left a comment.
My only concern is about the name. In every plugin we used `retry`+something and there `tries` which count the first call plus the retries in case of error.

I think it's more natural to use the `retries` so that a user expect certainly the first call and then if `retries > 0` other retry calls.

Put the first call plus retries in the same `tries` counter, is something that force me to read the guide to understand what it's meant for and go against the principle of least astonishment.

WDYT?

lib/logstash/inputs/elasticsearch.rb
kaisecheng: Yes, it catches `StandardError`. As the inputworker() fails to catch and restart plugin, any exception would crash Logstash which is the thing I want to avoid. Users running with many pipelines frustrate with the sudden shutdown. So I would say we need to have broader safeguard here.

lib/logstash/inputs/elasticsearch.rb
kaisecheng: I agree `retries` is a better word, more familiar to users and aligns the experience with other plugins. Thanks for calling a change!

kaisecheng: When I was writing `tries`, I have an impression that `retries` is confusing because I always need to check if it includes the first run. Probably, I am the only one who feels confused.

docs/index.asciidoc
```diff
@@ -424,6 +425,17 @@ When the `target` is set to a field reference, the `_source` of the hit is place
 This option can be useful to avoid populating unknown fields when a downstream schema such as ECS is enforced.
 It is also possible to target an entry in the event's metadata, which will be available during event processing but not exported to your outputs (e.g., `target \=> "[@metadata][_source]"`).
 
+
+[id="plugins-{type}s-{plugin}-retries"]
+===== `retries`
+
+* Value type is <<number,number>>
+* Default value is `1`
```
andsel, line 433: Is it correct that by default we force a retry or should it be 0 by default?

andsel (commented):
Thanks @kaisecheng  to have integrated the suggestions. I've left only a question/doubt.

docs/index.asciidoc
kaisecheng: good catch Andrea, it should be 0, no retry

andsel (approved):
LGTM

docs/index.asciidoc
```diff
@@ -424,6 +425,17 @@ When the `target` is set to a field reference, the `_source` of the hit is place
...
+[id="plugins-{type}s-{plugin}-retries"]
+===== `retries`
+
+* Value type is <<number,number>>
+* Default value is `0`
+
+The number of retries to run the query. If the query fails after all retries, it logs an error message.
+The default is 0, no retry. This value should be equal or greater than zero.
+
+
```
kaisecheng, line 438: @karenzone I have added `retries`option for the plugin. Please review the doc

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -262,22 +268,41 @@ def run(output_queue)
...
+    end
 
     logger.warn("managed slices for query is very large (#{@slices}); consider reducing") if @slices > 8
 
     pipeline_id = execution_context&.pipeline_id || 'main'
     @slices.times.map do |slice_id|
       Thread.new do
         LogStash::Util::set_thread_name("[#{pipeline_id}]|input|elasticsearch|slice_#{slice_id}")
-        do_run_slice(output_queue, slice_id)
+        retryable(JOB_NAME) do
```
mashhurs, line 286: QQ: with `slices` provided case, it seems we will face consistency issue here. Eg.: if we are dividing the work into 5 slices and first 3 slices fails with retry, then 2 succeeds. Maybe my ideation would not be a valid but IMHO, retrying seems not an ideal case for slices.

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -262,22 +268,41 @@ def run(output_queue)
...
       end
     end.map(&:join)
   end
 
+  def retryable(job_name, &block)
+    begin
+      stud_try = ::LogStash::Helpers::LoggableTry.new(logger, job_name)
+      stud_try.try((@retries + 1).times) {
+        block.call
+      }
```
yaauie, line 298: No need to create a new closure/binding. We can pass the block through.

```suggestion
      stud_try.try((@retries + 1).times, &block)
```

It used to be that `yield` was a bit faster than `block.call`, so perhaps this would be a more readable variant:

```suggestion
      stud_try.try((@retries + 1).times) { yield }
```

docs/index.asciidoc
```diff
@@ -124,6 +124,7 @@ This plugin supports the following configuration options plus the <<plugins-{typ
 | <<plugins-{type}s-{plugin}-ssl>> |<<boolean,boolean>>|No
 | <<plugins-{type}s-{plugin}-socket_timeout_seconds>> | <<number,number>>|No
 | <<plugins-{type}s-{plugin}-target>> | {logstash-ref}/field-references-deepdive.html[field reference] | No
+| <<plugins-{type}s-{plugin}-retries>> | <<number,number>>|No
```
karenzone, line 127: Should be in alphabetical order between `request_timeout` and `schedule`

docs/index.asciidoc
karenzone: Nice, thank you!

karenzone: Should be in alphabetical order between `request_timeout` and `schedule`

docs/index.asciidoc
```diff
@@ -424,6 +425,17 @@ When the `target` is set to a field reference, the `_source` of the hit is place
...
 It is also possible to target an entry in the event's metadata, which will be available during event processing but not exported to your outputs (e.g., `target \=> "[@metadata][_source]"`).
 
+
+[id="plugins-{type}s-{plugin}-retries"]
+===== `retries`
+
+* Value type is <<number,number>>
+* Default value is `0`
+
+The number of retries to run the query. If the query fails after all retries, it logs an error message.
```
karenzone, line 435:
```suggestion
The number of times to re-run the query after the first failure. If the query fails after all retries, it logs an error message.
```

karenzone: I'm trying to make the description different from the option name to help clarify for anybody who might be confused.  Let me know what you think about this, and if you'd like to tweak it some more.

docs/index.asciidoc
```diff
@@ -424,6 +425,17 @@ When the `target` is set to a field reference, the `_source` of the hit is place
...
+The default is 0, no retry. This value should be equal or greater than zero.
```
karenzone, line 436:
```suggestion
The default is 0 (no retry). This value should be equal to or greater than zero.
```

karenzone (commented):
I left some wording changes inline for consideration or discussion, and a request to put the new option in alphabetical order.    Let me know if you'd like to discuss. Otherwise, LGTM.

docs/index.asciidoc
kaisecheng: :facepalm: I renamed `tries` to `retries` but forgot to move it alphabetically

lib/logstash/inputs/elasticsearch.rb
kaisecheng: @mashhurs Thanks for raising the concern about data integrity. I changed the PR to discard the whole batch if one of the slices fails. Please review it.

mashhurs (approved):
🥇 LGTM and thanks for taking care of this.

PR #183, opened 2022-10-20
Tolerate failure

### LINKED ISSUE #182: Failure to create an event from the payload can crash the plugin
It is possible for Logstash event creation from a key/value map to fail with an exception (such as incorrect use of a reserved field in any version, or an invalid field reference as a key name prior to Logstash 8.3.0).

Instead of crashing, this plugin should catch the exception, re-serialize the offending map, and emit a tagged event (similar to how the JSON codec does so).

yaauie:
When an Event cannot be created directly from the hit, or when the
docinfo cannot be merged into a non-hash field in the hit, emit an
Event tagged with `_elasticsearch_input_failure` that contains the
JSON-encoded hit in `[event][original]` instead of crashing.

Resolves #182


docs/index.asciidoc
```diff
@@ -93,6 +93,16 @@ The plugin logs a warning when ECS is enabled and `target` isn't set.
...
 TIP: Set the `target` option to avoid potential schema conflicts.
 
+[id="plugins-{type}s-{plugin}-failure-handling"]
+==== Failure handling
+
+When this input plugin cannot create a structured `Event` from a hit result, it will instead create an `Event` that is tagged with `_elasticsearch_input_failure` whose `[event][original]` is a JSON-encoded string representation of the entire hit.
+
+Common causes are:
+
+ - When the hit result contains top-level fields that are {logstash-ref}/processing.html#reserved-fields[reserved in Logstash] but do not have the expected shape. Use the <<plugins-{type}s-{plugin}-target>> directive to avoid conflicts with the top-level namespace.
```
yaauie, line 103: @karenzone can I get :eyes: on this one to make sure I'm not risking breaking the docs build?

It looks like we added `{logstash-ref}/processing.html#reserved-fields` in 7.7 (it doesn't exist on 7.6 and before), but is present on all of the versions of the reference available in the quick picker (`master`, `current`, `8.5`, `8.4`, `7.17`).

karenzone, line 103:
```suggestion
 - The hit result contains top-level fields that are {logstash-ref}/processing.html#reserved-fields[reserved in Logstash], but do not have the expected shape. Use the <<plugins-{type}s-{plugin}-target>> directive to avoid conflicts with the top-level namespace.
```

docs/index.asciidoc
```diff
@@ -93,6 +93,16 @@ The plugin logs a warning when ECS is enabled and `target` isn't set.
...
+ - When <<plugins-{type}s-{plugin}-docinfo>> is enabled and the docinfo fields cannot be merged into the hit result. Combine <<plugins-{type}s-{plugin}-target>> and <<plugins-{type}s-{plugin}-docinfo_target>> to avoid conflict.
```
karenzone, line 104:
```suggestion
 - The <<plugins-{type}s-{plugin}-docinfo>> option is enabled, and the docinfo fields cannot be merged into the hit result. Combine <<plugins-{type}s-{plugin}-target>> and <<plugins-{type}s-{plugin}-docinfo_target>> to avoid conflict.
```

docs/index.asciidoc
karenzone: > It looks like we added {logstash-ref}/processing.html#reserved-fields in 7.7 (it doesn't exist on 7.6 and before), but is present on all of the versions of the reference available in the quick picker (master, current, 8.5, 8.4, 7.17).

The gemlock release file controls which plugin versions get mapped/picked up for each stack version when we run docgen. For that reason, the plugin<->stack versioning generally works itself out without any manual intervention from us.

karenzone (commented):
Left some minor wording tweaks inline for your consideration and addressed versioning question.  Otherwise, LGTM

TL;DR. The versioning should work itself out and your linking in spot on.

karenzone (approved):
Signing off on the docs portion in case you're waiting for an approval to move forward.

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -350,24 +350,42 @@ def run(output_queue)
   end
 
   ##
+  # @param output_queue [#<<]
+  # @param scroll_id [String]: a scroll id to resume
+  # @return [Array(Boolean,String)]: a tuple representing whether the response
+  #
+  def process_next_scroll(output_queue, scroll_id)
```
donoghuc, line 357: I cant quite figure out how this is used.

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -350,24 +350,42 @@ def run(output_queue)
...
 
+  def event_from_hit(hit, root_field)
+    event = targeted_event_factory.new_event hit[root_field]
+    set_docinfo_fields(hit, event) if @docinfo
+
+    event
+  rescue => e
+    serialized_hit = hit.to_json
+    logger.warn("Event creation error, original data now in [event][original] field", message: e.message, exception: e.class, data: serialized_hit)
+    return event_factory.new_event('event' => { 'original' => serialized_hit }, 'tags' => ['_elasticsearch_input_failure'])
```
donoghuc, line 379:
```suggestion
   event_factory.new_event('event' => { 'original' => serialized_hit }, 'tags' => ['_elasticsearch_input_failure'])
```
Nit: Shouldnt need the explicit return, but I see that is not the only instance in this project.

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -350,24 +350,42 @@ def run(output_queue)
...
     decorate(event)
     output_queue << event
   end
 
+  def event_from_hit(hit, root_field)
+    event = targeted_event_factory.new_event hit[root_field]
+    set_docinfo_fields(hit, event) if @docinfo
+
+    event
+  rescue => e
```
donoghuc, line 376:
```suggestion
  rescue RuntimeError => e
```
Looks like `set_docinfo_fields` explicitly raises a `RuntimeError` now? May be best to be as specific as possible with error catching.

spec/fixtures/test_certs/renew.sh
```diff
@@ -0,0 +1,15 @@
...
+openssl x509 -x509toreq -in ca.crt -copy_extensions copyall -signkey ca.key -out ca.csr
+openssl x509 -req -copy_extensions copyall -days 365 -in ca.csr -set_serial 0x01 -signkey ca.key -out ca.crt && rm ca.csr
+openssl x509 -in ca.crt -outform der | sha256sum | awk '{print $1}' > ca.der.sha256
+
+openssl x509 -x509toreq -in es.crt -copy_extensions copyall -signkey es.key -out es.csr
+openssl x509 -req -copy_extensions copyall -days 365 -in es.csr -set_serial 0x01 -CA ca.crt -CAkey ca.key -out es.crt && rm es.csr
+cat es.crt ca.crt > es.chain.crt
+
+# output ISO8601 timestamp to file
+date -Iseconds > GENERATED_AT
```
donoghuc, line 15: I feel like expiration date would be more useful 🤷 or both. 
```
openssl x509 -enddate -noout -in es.crt | sed 's/notAfter=//'
```

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -350,24 +350,42 @@ def run(output_queue)
...
+    serialized_hit = hit.to_json
```
donoghuc, line 377: I'm assuming if these were not json serializable in the first place we would never get this far?

donoghuc (commented):
Test cleanup looks good (one small suggestion on cert generation script). 

Release prep looks correct (version number is good and changelog looks accurate)

Had a couple nit comments and a question about potential unused code. Other than that this looks like a good improvement.

lib/logstash/inputs/elasticsearch.rb
yaauie: It is technically redundant, but present for readability.

The explicit `return` statement makes clear to me that we are _intentionally_ returning that object from the `rescue`... block, because otherwise when I'm reading code I don't know whether the return value matters to the caller(s).

It's a side-effect of using exceptions for flow-control (which is less than ideal, but it is what we are already starting with).

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -350,24 +350,42 @@ def run(output_queue)
...
+  # @param output_queue [#<<]
+  # @param scroll_id [String]: a scroll id to resume
+  # @return [Array(Boolean,String)]: a tuple representing whether the response
+  #
+  def process_next_scroll(output_queue, scroll_id)
+    r = scroll_request(scroll_id)
+    r['hits']['hits'].each { |hit| push_hit(hit, output_queue) }
+    [r['hits']['hits'].any?, r['_scroll_id']]
+  end
+
```
yaauie, line 362: This was a remnant from before I rebased, and will ✂️ snip.

```suggestion
```

lib/logstash/inputs/elasticsearch.rb
yaauie: Yes, but this isn't the only exception we're handling here.

We know that `EventFactory#new_event` (which delegates to `Event::new`) can give us exceptions too, specifically when the payload uses reserved fields in incorrect ways. We want to catch any situation in which we fail to create an event from the hit.

lib/logstash/inputs/elasticsearch.rb
yaauie: Ehhh, we deserialized the hit _from_ json, so it should be able to go back?

If we can't, we are at least not creating a _new_ failure scenario (merely failing to handle something that currently already crashes the plugin).

lib/logstash/inputs/elasticsearch.rb
donoghuc: I like that pattern. 👍

lib/logstash/inputs/elasticsearch.rb
donoghuc: I see, Ok yeah that makes sense.

donoghuc (approved):
LGTM. Ready release and forward port to `main` branch.

Issue #188, opened 2023-04-24
Results from all scrolls for each slice are stored in memory leading to OOM


jsvd:
In [PR#179](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/179), a [temporary array](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/179/files#diff-04cb591df2446500a108eebbfe57c8ed0c72553d33abe1bcef68a5a15474b097R325-R327) is used to store the results for each slice.
For each slice, the multiple scrolls are executed and stored in the array, leading to OOM.

kaisecheng:
this issue is fixed in `4.17.1`

PR #189, opened 2023-04-25
fix scroll slice memory

kaisecheng:
Fixed: #188

The changes in https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/179 cause each of the slices to collect all data from all scrolls into a temporary array (https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/179/files#diff-04cb591df2446500a108eebbfe57c8ed0c72553d33abe1bcef68a5a15474b097R325), which causes a build up of memory, while before this change each scroll on each slice would be pushed to the queue and freed before proceeding to the next.

In this PR, the successful slices are pushed to the queue immediately without waiting for the other slices. Failed slice raises an exception and gets retries based on the `retries` setting. If the failed slice has already indexed some data, the retries could produce duplicated data

### DIFF
--- a/spec/inputs/elasticsearch_spec.rb
+++ b/spec/inputs/elasticsearch_spec.rb
@@ -414,18 +414,19 @@ describe LogStash::Inputs::Elasticsearch, :ecs_compatibility_support do
           expect(Elasticsearch::Client).to receive(:new).with(any_args).and_return(client)
           plugin.register
 
-          expect(client).to receive(:clear_scroll).and_return(nil)
+          expect(client).to receive(:clear_scroll).twice.and_return(nil)
 
-          # SLICE0 is a three-page scroll in which the second page throw exception
+          # SLICE0 is a three-page scroll
           slice0_query = LogStash::Json.dump(query.merge('slice' => { 'id' => 0, 'max' => 2}))
           expect(client).to receive(:search).with(hash_including(:body => slice0_query)).and_return(slice0_response0)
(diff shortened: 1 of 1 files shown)


spec/inputs/elasticsearch_spec.rb
```diff
@@ -408,45 +408,6 @@ def synchronize_method!(object, method_name)
           end
         end
       end
-
```
roaksoax, line 411: Do the currently available tests cover use case you are fixing by this PR already? If not, lets add tests that cover that scenario specifically and ensure it wont regress if changed.

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -326,63 +326,37 @@ def run(output_queue)
...
     logger.warn("managed slices for query is very large (#{@slices}); consider reducing") if @slices > 8
 
-    slice_results = parallel_slice # array of tuple(ok, events)
-
-    # insert events to queue if all slices success
-    if slice_results.all?(&:first)
-      slice_results.flat_map { |success, events| events }
-                  .each { |event| output_queue << event }
-    end
+    pipeline_id = execution_context&.pipeline_id || 'main'
```
jsvd, line 336: this can be moved to register

jsvd (commented):
Tested and works as expected.
I'd be nice to include a test that confirms that:
1. we call << on queue as many times as docs in the all scrolls of each slice
2. we don't wait until the last scroll to write to the queue

maybe for this we can check that by the first scroll we have called << equal to the page size?

PR #193, opened 2023-05-31
fix: restore use of DefaultHostnameVerifier

edmocosta:
The legacy and out-of-date StrictHostnameVerifier from Apache HTTP Client is deprecated with guidance to use the DefaultHostnameVerifier, which is in fact a strict but modern verifier that can handle modern identity claims.

Resolves: https://github.com/logstash-plugins/logstash-input-elasticsearch/issues/192


jsvd (approved):
LGTM

PR #195, opened 2023-09-04
Add Elastic Api Version header

kaisecheng:
This commit adds "Elastic-Api-Version" : "2023-10-31" to request header when the Elasticsearch is serverless

### DIFF
--- a/lib/logstash/inputs/elasticsearch.rb
+++ b/lib/logstash/inputs/elasticsearch.rb
@@ -687,7 +687,7 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
   end
 
   def serverless?
-    build_flavor == BUILD_FLAVOR_SERVERLESS
+    @is_serverless ||= (build_flavor == BUILD_FLAVOR_SERVERLESS)
   end
 
   def add_headers(options={})
--- a/lib/logstash/inputs/elasticsearch.rb
+++ b/lib/logstash/inputs/elasticsearch.rb
@@ -680,12 +680,10 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
   end
 
   def test_serverless_connection!
-    begin
-      @client.info(:headers => DEFAULT_EAV_HEADER ) if serverless?
-    rescue => e
-      @logger.error("Failed to retrieve Elasticsearch info", message: e.message, exception: e.class, backtrace: e.backtrace)
-      raise LogStash::ConfigurationError, "Could not connect to a compatible version of Elasticsearch"
-    end
+    @client.info(:headers => DEFAULT_EAV_HEADER ) if serverless?
+  rescue => e
+    @logger.error("Failed to retrieve Elasticsearch info", message: e.message, exception: e.class, backtrace: e.backtrace)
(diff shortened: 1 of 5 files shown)


jsvd (changes requested):
inspecting the request done by the manticore adapter in the ES ruby client, we can see the header is not being sent to the server.
a simple test is to bump the date to 2024 and confirm that we can still query a serverless deployment.
My suspicion is around header handling between the headers passed during client creation vs headers passed at request time.

kaisecheng:
Good catch! The client got headers but completely ignored and [overwritten](https://github.com/elastic/elastic-transport-ruby/blob/main/lib/elastic/transport/transport/http/manticore.rb#L93C31-L93C47) by initial headers. It seems a bug only in manticore implementation as [faraday](https://github.com/elastic/elastic-transport-ruby/blob/main/lib/elastic/transport/transport/http/faraday.rb#L39C43-L39C43) and [curb](https://github.com/elastic/elastic-transport-ruby/blob/main/lib/elastic/transport/transport/http/curb.rb#L37C39-L37C39) do a merge with the input headers.

kaisecheng:
link the upstream issue https://github.com/elastic/elastic-transport-ruby/issues/66

kaisecheng:
upstream has merged the fix. We are now waiting for the release of new version

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -668,6 +679,31 @@ def test_connection!
...
   end
 
+  def test_serverless_connection!
+    begin
+      @client.info(:headers => DEFAULT_EAV_HEADER ) if serverless?
+    rescue => e
+      @logger.error("Failed to retrieve Elasticsearch info", message: e.message, exception: e.class, backtrace: e.backtrace)
+      raise LogStash::ConfigurationError, "Could not connect to a compatible version of Elasticsearch"
+    end
+  end
```
jsvd, line 689: in these cases we can get rid of the level of indentation and just do:

```suggestion
  def test_serverless_connection!
    @client.info(:headers => DEFAULT_EAV_HEADER ) if serverless?
  rescue => e
    @logger.error("Failed to retrieve Elasticsearch info", message: e.message, exception: e.class, backtrace: e.backtrace)
    raise LogStash::ConfigurationError, "Could not connect to a compatible version of Elasticsearch"
  end
```

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -668,6 +679,31 @@ def test_connection!
...
+
+  def cluster_info
+    @client.info
+  end
+  def build_flavor
```
jsvd, line 694:
```suggestion
  end

  def build_flavor
```

jsvd (approved):
minor code style fixes. Tested and works great. Would love to not have to have this add_headers constantly, but requires deeper refactoring.
LGTM

kaisecheng:
@jsvd The last commit recreated the client with header. es-input should be straightforward to replace the client

jsvd (approved):
LGTM

kaisecheng:
link the issue for cleaning up resources https://github.com/logstash-plugins/logstash-input-elasticsearch/issues/196

PR #90, opened 2018-02-15
Adding support to return aggregations

aclowkey:
Currently there's only support for getting hits from ElasticSearch's response. This is probably the most common use case, however, sometimes we might want to get the aggregations information.

This pull request is to add a new parameter _response_type_ which allows you to specify whether you need the hits, or aggregations and returns the relevant data.

PR #197, opened 2023-10-11
Enable aggregation queries

### LINKED ISSUE #58: Support Aggregations
Not sure if this would be better supported by a totally different plugin, but I think it would make sense to support aggregation results.  

As an example, say we wanted to use a sum aggregation, by terms.

```
GET myserverlogs*/_search
{

   "size": 0,
   "aggs": {
      "types": {
         "terms": {
            "field": "hostname"
         },
         "aggs": {
            "connections": {
               "sum": {
                  "field": "connections"
               }
            }
         }
      }
   }
}
```

Results:
```
  "aggregations": {
      "types": {
         "doc_count_error_upper_bound": 0,
         "sum_other_doc_count": 0,
         "buckets": [
            {
               "key": "hostA",
               "doc_count": 90310,
               "total_connections": {
                  "value": 2344
               }
            },
            {
               "key": "hostB",
               "doc_count": 485,
               "total_connections": {
                  "value": 233
               }
            },
            {
               "key": "hostC",
               "doc_count": 485,
               "total_connections": {
                  "value": 123
               }
            }
         ]
      }
   }
```

Then in Logstash we could say something like:

```
input {
  elasticsearch {
    ...
    results_mode => "aggregation"
    source_array => "types.buckets"
  }
}
```

This could even be used to create aggregation based summary indices on a regular interval.  I remember @polyfractal was working on an ES plugin that did something similar last year.

MikhailMS:
As title says, this PR enables this plugin to execute aggregation queries

The work is based on the existing PR #90 (and potentially resolves #58)
However code & tests were updated to include all the changes that were introduced since original PR was created/last updated

Tests are included and also was live tested with Logstash v8.8.2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## 4.19.0
-  - Added `response_type` configuration option to allow aggregations to be executed:w
+  - Added `response_type` configuration option to allow aggregations to be executed
 
 ## 4.18.0
   - Added request header `Elastic-Api-Version` for serverless [#195](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/195)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 4.19.0
+  - Added `response_type` configuration option to allow aggregations to be executed:w
+
 ## 4.18.0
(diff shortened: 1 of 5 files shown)


jsvd:
replaced by #202

jsvd:
Thanks @MikhailMS and @aclowkey for the work, this has been committed in #202 (I retained co-attribution on the squashed commit).
This has been published in version [4.20.0](https://rubygems.org/gems/logstash-input-elasticsearch/versions/4.20.0).

PR #198, opened 2023-10-27
Add support to search after

kaisecheng:
This commit introduces the `search_api` option to enable support for both `search_after` and `scroll`. 

| `search_api`  | description  |
|-----------|-----------|
| `auto`    | Default value. It uses `search_after` for Elasticsearch 8 and newer versions, otherwise, it falls back to the `scroll` API    |
| `search_after`    | Create a point in time for search, which is the recommended way to do paginated search since 7.10   |
| `scroll`   | Uses scroll API to search, which is the previous way to search    |

When `search_after` is utilized, if the query doesn't specify a sort field, a default sort of `{ "sort": { "_shard_doc": "asc" } }` will be added to the query. By default, Elasticsearch adds a sort field "_shard_doc" on top of existing sort implicitly as a tie-breakers.

The scroll search has been refactored and is expected to function the same, with only minor changes made to the logging messages.

### Backward compatibility
There is one concern to make search_after as the default. The scroll API doesn't require a sort field in query, while search_after needs the last sort value from the return documents for the next search. When the plugin is updated, if the query has sort, it is fine as both APIs should return data in the same order. If the query doesn't has sort, I believe user doesn't concern the ordering, hence, adding a default sort field "_shard_doc" should be fine.

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -390,7 +390,7 @@ round trip (i.e. between the previous scroll request, to the next).
 `auto` uses `search_after` for Elasticsearch version `8.0.0` or higher, otherwise, fall back to `scroll` API.
 
 `search_after` uses {ref}/point-in-time-api.html#point-in-time-api[point in time] and sort value to search.
-The query requires to have at least one `sort` field. Check out <<plugins-{type}s-{plugin}-query>>.
+The query requires at least one `sort` field, as described in the <<plugins-{type}s-{plugin}-query>> parameter.
 
 `scroll` uses {ref}/paginate-search-results.html#scroll-search-results[scroll] API to search, which is no longer recommended.
 
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -387,7 +387,7 @@ round trip (i.e. between the previous scroll request, to the next).
 * Value can be any of: `auto`, `search_after`, `scroll`
 * Default value is `auto`
 
-With `auto` the plugin uses the `search_after` parameter for Elasticsearch version `8.0.0` or higher, otherwise, the `scroll` API is used instead.
+With `auto` the plugin uses the `search_after` parameter for Elasticsearch version `8.0.0` or higher, otherwise the `scroll` API is used instead.
 
 `search_after` uses {ref}/point-in-time-api.html#point-in-time-api[point in time] and sort value to search.
 The query requires at least one `sort` field, as described in the <<plugins-{type}s-{plugin}-query>> parameter.
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -335,7 +335,7 @@ The query to be executed. Read the {ref}/query-dsl.html[Elasticsearch query DSL
 documentation] for more information.
 
 When <<plugins-{type}s-{plugin}-search_api>> resolves to `search_after` and the query does not specify `sort`,
-the default sort `'{ "sort": { "_shard_doc": "asc" } }'` will be added to the query. Check out {ref}/paginate-search-results.html#search-after[Elasticsearch search after]
+the default sort `'{ "sort": { "_shard_doc": "asc" } }'` will be added to the query. Please refer to the {ref}/paginate-search-results.html#search-after[Elasticsearch search_after] parameter to know more.
 
 [id="plugins-{type}s-{plugin}-request_timeout_seconds"]
 ===== `request_timeout_seconds`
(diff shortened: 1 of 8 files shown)


lib/logstash/inputs/elasticsearch/paginated_search.rb
```diff
@@ -0,0 +1,237 @@
...
+          raise NotImplementedError
+        end
+
+        def retryable_slice_search(output_queue)
+          raise NotImplementedError
+        end
+      end
+
+      class Scroll < PaginatedSearch
+        SCROLL_JOB = "run scroll"
```
jsvd, line 51:
```suggestion
        SCROLL_JOB = "scroll paginated search"
```

lib/logstash/inputs/elasticsearch/paginated_search.rb
```diff
@@ -0,0 +1,237 @@
...
+          @client.clear_scroll(:body => { :scroll_id => scroll_id }) if scroll_id
+        rescue => e
+          # ignore & log any clear_scroll errors
+          logger.warn("Ignoring clear_scroll exception", message: e.message, exception: e.class)
+        end
+      end
+
+      class SearchAfter < PaginatedSearch
+        PIT_JOB = "create point in time"
+        SEARCH_AFTER_JOB = "run search after"
```
jsvd, line 127:
```suggestion
        SEARCH_AFTER_JOB = "search_after paginated search"
```

lib/logstash/inputs/elasticsearch/paginated_search.rb
```diff
@@ -0,0 +1,237 @@
...
+        def clear(scroll_id)
+          @client.clear_scroll(:body => { :scroll_id => scroll_id }) if scroll_id
+        rescue => e
+          # ignore & log any clear_scroll errors
+          logger.warn("Ignoring clear_scroll exception", message: e.message, exception: e.class)
+        end
+      end
+
+      class SearchAfter < PaginatedSearch
+        PIT_JOB = "create point in time"
```
jsvd, line 126:
```suggestion
        PIT_JOB = "create point in time (PIT)"
```

lib/logstash/inputs/elasticsearch/paginated_search.rb
```diff
@@ -0,0 +1,237 @@
...
+        SEARCH_AFTER_JOB = "run search after"
+
+        def pit?(id)
+          !!id&.is_a?(String)
+        end
+
+        def create_pit
+          logger.info("create point in time")
```
jsvd, line 134:
```suggestion
          logger.info("Create point in time (PIT)")
```

lib/logstash/inputs/elasticsearch/paginated_search.rb
```diff
@@ -0,0 +1,237 @@
...
+                end
+              end
+            end.map(&:join)
+          end
+
+          logger.trace("#{@slices} slices completed")
+        end
+
+        def clear(pit_id)
+          logger.info("close point in time")
```
jsvd, line 228:
```suggestion
          logger.info("Closing point in time (PIT)")
```

lib/logstash/inputs/elasticsearch/paginated_search.rb
```diff
@@ -0,0 +1,237 @@
...
+          @client.close_point_in_time(:body => {:id => pit_id} ) if pit?(pit_id)
+        rescue => e
+          logger.warn("Ignoring close_point_in_time exception", message: e.message, exception: e.class)
```
jsvd, line 231: If there's nothing the user can do here and we're just closing the PIT, we can make this a debug entry instead:

```suggestion
          logger.debug("Ignoring close_point_in_time exception", message: e.message, exception: e.class)
```

docs/index.asciidoc
```diff
@@ -377,6 +381,19 @@ This parameter controls the keepalive time in seconds of the scrolling
 request and initiates the scrolling process. The timeout applies per
 round trip (i.e. between the previous scroll request, to the next).
 
+[id="plugins-{type}s-{plugin}-seearch_api"]
+===== `search_api`
+
+* Value can be any of: `auto`, `search_after`, `scroll`
+* Default value is `auto`
+
+`auto` uses `search_after` for Elasticsearch version `8.0.0` or higher, otherwise, fall back to `scroll` API.
```
jsvd, line 390:
```suggestion
With `auto` the plugin uses the `search_after` parameter for Elasticsearch version `8.0.0` or higher, otherwise, the `scroll` API is used instead.
```

docs/index.asciidoc
```diff
@@ -377,6 +381,19 @@ This parameter controls the keepalive time in seconds of the scrolling
...
+
+`search_after` uses {ref}/point-in-time-api.html#point-in-time-api[point in time] and sort value to search.
+The query requires to have at least one `sort` field. Check out <<plugins-{type}s-{plugin}-query>>.
```
jsvd, line 393:
```suggestion
The query requires at least one `sort` field, as described in the <<plugins-{type}s-{plugin}-query>> parameter.
```

jsvd (commented):
Overall the feature works as expected, I've tested both modes, with and without slices.
Added a few tweaks to code structure and error messages

jsvd (approved):
Minor tweaks to Docs, otherwise LGTM

PR #200, opened 2023-12-04
Bump to v4.19.1 to pick up doc fix needed for docgen

karenzone:
Some doc formatting errors don't require version bumps, and can wait until the next time we need to bump versions.  Turns out that the fix in #199 does require a version bump to get us past CI failures.  I can't fix the error in output (as I did for the VPR) until we successfully generate output.

Screen Shot 2023-12-04 at 3 59 25 PM


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 4.19.1
+  - Plugin version bump to pick up docs fix in  [#199](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/199 required to clear build error in docgen. [#200](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/200)
```
karenzone, line 2:
```suggestion
  - Plugin version bump to pick up docs fix in  [#199](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/199) required to clear build error in docgen. [#200](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/200)
```

PR #202, opened 2024-01-22
Support aggregations

### LINKED ISSUE #58: Support Aggregations

jsvd:
Add `response_type` configuration option to allow processing result of aggregations.

The default `hits` will generate one event per returned document (i.e. "hit"), which is the current behavior.

When set to `aggregations`, a single Logstash event will be generated with the contents of the `aggregations` object of the query's response. In this case the `hits` object will be ignored.
The parameter `size` will be always be set to 0 regardless of the default or user-defined value set in this plugin.

builds upon and replaces #197 
closes #58

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 4.20.0
+  - Added `response_type` configuration option to allow aggregations to be executed:w
+
 ## 4.19.1
   - Plugin version bump to pick up docs fix in  [#199](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/199) required to clear build error in docgen. [#200](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/200)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## 4.20.0
-  - Added `response_type` configuration option to allow aggregations to be executed
+  - Added `response_type` configuration option to allow processing result of aggregations
 
 ## 4.19.1
   - Plugin version bump to pick up docs fix in  [#199](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/199) required to clear build error in docgen. [#200](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/200)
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 4.20.0
+  - Added `response_type` configuration option to allow aggregations to be executed
```
jsvd, line 2:
```suggestion
  - Added `response_type` configuration option to allow processing result of aggregations
```

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 4.20.0
+  - Added `response_type` configuration option to allow processing result of aggregations
```
jsvd, line 2:
```suggestion
  - Added `response_type` configuration option to allow processing result of aggregations [#202](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/202)
```

PR #204, opened 2024-03-04
fix case when aggregation returns an error

jsvd:


jsvd:
@logstashmachine bump patch

kaisecheng (approved):
LGTM

PR #205, opened 2024-05-29
Introduce cursor tracking akin to jdbc input

### LINKED ISSUE #93: [feature] Support configurations so it behaves like JDBC input
Currently, the ES input plugin has no way to pick up from where it left off last time it executed, just as the JDBC input does with the `sql_last_value` value.

I propose a set of configurations that makes ES input plugin query from a past known `@timestamp` (by default)

### LINKED ISSUE #184: Add a "sincedb" type of mecanism
Hi,

Would it be possible to add a "sincedb" type of functionality to this plugin ? 

You can schedule a query to pull logs from Elastic at regular intervals for example 1 minute and configure the query to return 1 minute's worth of logs. The issue is that if a crash occurs or if the logstash is restarted, logs will be missing. The issue arises only when trying to "stream" events from elastic into logstash continuously. Other plugins such as the JDBC plugin let you configure a sincedb to avoid these issues.

Is there another way to reliably stream logs from elastic to logstash ?

Thanks

jsvd:
With this change it's possible to configure a "tracking_field" whose value is injected into the query using the `:last_value` keyword, and a `:present` field that represents `now - 30s`.

Example configuration:

```
input {
  elasticsearch {
    id => tail
    hosts => [ 'https://host...']
    api_key => "..."
    index => "my_data"
    query => '{ "query": { "range": { "event.ingested": { "gt": ":last_value", "lt": ":present"}}}, "sort": [ { "event.ingested": {"order": "asc", "format": "strict_date_optional_time_nanos", "numeric_type" : "date_nanos" } } ] }'
    tracking_field => "[event][ingested]"
    tracking_field_seed => "1980-01-01T23:59:59.999999999Z"
    slices => 5
    schedule => '* * * * *'
    schedule_overlap => false
  }
}
output {
  elasticsearch { .. }
}
```

To test this, follow the guidance on the doc changes from this PR on how to set up Elasticsearch. Or you can use a script that prepares the index/ingest_pipeline and writes documents to ES like https://gist.github.com/jsvd/6ea96197078d906362227494e0c542dc

resolves #184 
resolves #93

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -217,7 +217,7 @@ added by an Elasticsearch ingest pipeline, and the `tracking_field` capability o
 With this setup, as new documents are indexed an `test-*` index, the next scheduled run will:
 
 . select all new documents since the last observed value of the tracking field;
-. use <<point-in-time-api,Point in time (PIT)>> + <<search-after, Search after>> to paginate through all the data;
+. use {ref}/point-in-time-api.html#point-in-time-api[Point in time (PIT)] + {ref}/paginate-search-results.html#search-after[Search after] to paginate through all the data;
 . update the value of the field at the end of the pagination.
 
 [id="plugins-{type}s-{plugin}-options"]
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -106,7 +106,6 @@ Common causes are:
 [id="plugins-{type}s-{plugin}-cursor"]
 ==== Tracking a field's value across runs
 
-NOTE: experimental:[] `tracking_field` and related settings are experimental and subject to change in the future
 .Technical Preview: Tracking a field's value
 ****
 The feature that allows tracking a field's value across runs is in _Technical Preview_.
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -49,6 +49,119 @@ This would create an Elasticsearch query with the following format:
     }'
 
 
+[id="plugins-{type}s-{plugin}-cursor"]
+==== Tracking a field's value across runs
+
+It is sometimes desirable to track the value of a particular field between two jobs:
+* avoid re-processing the entire result set of a long query after an unplanned restart
+* only grab new data from an index instead of processing the entire set on each job
+
+For this, the Elasticsearch input plugin provides the <<tracking_field>> and <<tracking_field_seed>> options.
+When <<tracking_field>> is set, the plugin will record the value of that field for the last document retrieved in a run into
+a file (location defaults to <<last_run_metadata_path>>.
+
+The user can then inject this value in the query using the placeholder `:last_value`. The value will be injected into the query
+before execution, and the updated after the query completes, assuming new data was found.
+
+The plugin also offers another placeholder called `:present` used to inject the nano-second based
+
+This feature works best when:
+* the query sorts by the tracking field
+* the field type has enough resolution so that two events are unlikely to have the same value for the field
+
+A suggestion is to use a tracking field that has nanosecond second precision, like
+https://www.elastic.co/guide/en/elasticsearch/reference/current/date_nanos.html[date nanoseconds] field type.
+
+A good use case for this feature is to track new data in an index, which can be achieved by:
+
+1. create ingest pipeline that adds Elasticsearch's `_ingest.timestamp` field to the documents as `event.ingested`:
+
+[source, json]
+    PUT _ingest/pipeline/my-pipeline
+    {
+      "processors": [
+              {
+            "script": {
+              "lang": "painless",
+              "source": "ctx.putIfAbsent(\"event\", [:]); ctx.event.ingested = metadata().now.format(DateTimeFormatter.ISO_INSTANT);"
+            }
+          }
+      ]
+    }
+
+
+2. create an index mapping where the tracking field is of date nanosecond type and invokes the defined pipeline:
+
+[source, json]
+    PUT /_template/my_template
+    {
+      "index_patterns": ["test-*"],
+      "settings": {
+        "index.default_pipeline": "my-pipeline",
+      },
+      "mappings": {
+        "properties": {
+          "event": {
+            "properties": {
+              "ingested": {
+                "type": "date_nanos",
+                "format": "strict_date_optional_time_nanos"
+              }
+            }
+          }
+        }
+      }
+    }
+
+3. define a query that looks at all data of the indices, sorted by the tracking field, and with a range filter since the last value seen until present:
+
+[source,json]
+{
+  "query": {
+    "range": {
+      "event.ingested": {
+        "gt": ":last_value",
+        "lt": ":present"
+      }
+    }
+  },
+  "sort": [
+    {
+      "event.ingested": {
+        "order": "asc",
+        "format": "strict_date_optional_time_nanos",
+        "numeric_type": "date_nanos"
+      }
+    }
+  ]
+}
+
+4. configure the Elasticsearch input to query the indices with the query defined above, every minute, and track the `event.ingested` field:
+
+[source, ruby]
+    input {
+      elasticsearch {
+        id => tail_test_index
+        hosts => [ 'https://..']
+        api_key => '....'
+        index => 'test-*'
+        query => '{ "query": { "range": { "event.ingested": { "gt": ":last_value", "lt": ":present"}}}, "sort": [ { "event.ingested": {"order": "asc", "format": "strict_date_optional_time_nanos", "numeric_type" : "date_nanos" } } ] }'
+        tracking_field => "[event][ingested]"
+        # set a seed value to a value known to be older than any value of `event.ingested`
+        tracking_field_seed => "1980-01-01T23:59:59.999999999Z"
+        slices => 5 # optional use of slices to speed data processing, should be less than number of primary shards
+        schedule => '* * * * *' # every minute
+        schedule_overlap => false # don't accumulate jobs if one takes longer than 1 minute
+      }
+    }
+
+With this setup, as new documents are indexed an `test-*` index, the next scheduled run will:
+
+1. select all new documents since the last observed value of the tracking field;
+2. use PIT+search_after to paginate through all the data;
+3. update the value of the field at the end of the pagination.
+
+[id="plugins-{type}s-{plugin}-scheduling"]
 ==== Scheduling
 
 Input from this plugin can be scheduled to run periodically according to a specific
@@ -658,6 +771,8 @@ Which field from the last event of a previous run will be used a cursor value fo
 The value of this field is injected into each query if the query uses the placeholder `:last_value`.
 For the first query after a pipeline is started, the value used is either read from <<last_run_metadata_path>> file,
 or taken from <<tracking_field_seed>> setting.
+
+Note: The tracking value is updated only after the PIT+search_after run completes, it won't update during the search_after pagination. This is to allow use of slices.
 
 [id="plugins-{type}s-{plugin}-tracking_field_seed"]
 ===== `tracking_field_seed`
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -454,7 +454,9 @@ referencing multiple indices.
   * There is no default value for this setting.
 
 The path to store the last observed value of the tracking field, when used.
-By default this file is stored as `<path.data>/plugins/inputs/elasticsearch/<pipeline.id>.last_run`.
+By default this file is stored as `<path.data>/plugins/inputs/elasticsearch/last_run_value`.
+
+This setting should point to file, not a directory, and Logstash must have read+write access to this file.
 
 [id="plugins-{type}s-{plugin}-password"]
 ===== `password`
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -117,18 +117,21 @@ When <<plugins-{type}s-{plugin}-tracking_field>> is set, the plugin will record 
 a file (location defaults to <<plugins-{type}s-{plugin}-last_run_metadata_path>>).
 
 The user can then inject this value in the query using the placeholder `:last_value`. The value will be injected into the query
-before execution, and the updated after the query completes, assuming new data was found.
+before execution, and the updated after the query completes if new data was found.
 
 This feature works best when:
-* the query sorts by the tracking field
-* the field type has enough resolution so that two events are unlikely to have the same value for the field
 
-The plugin also offers another placeholder called `:present` used to inject the nano-second based value of "now-30s".
+. the query sorts by the tracking field;
+. the timestamp field is added by {es};
+. the field type has enough resolution so that two events are unlikely to have the same value.
 
-A suggestion is to use a tracking field that has nanosecond second precision, like
-https://www.elastic.co/guide/en/elasticsearch/reference/current/date_nanos.html[date nanoseconds] field type.
+It is recommended to use a tracking field whose type is https://www.elastic.co/guide/en/elasticsearch/reference/current/date_nanos.html[date nanoseconds].
+If the tracking field is of this data type, an extra placeholder called `:present` can be used to inject the nano-second based value of "now-30s".
+This placeholder is useful as the right-hand side of a range filter, allowing the collection of
+new data but leaving partially-searcheable bulk request data to the next scheduled job.
 
-A good use case for this feature is to track new data in an index, which can be achieved by:
+Below is a series of steps to help set up the "tailing" of data being written to a set of indices, using a date nanosecond field
+added by an Elasticsearch ingest pipeline, and the `tracking_field` capability of this plugin.
 
 . create ingest pipeline that adds Elasticsearch's `_ingest.timestamp` field to the documents as `event.ingested`:
 
@@ -205,9 +208,7 @@ A good use case for this feature is to track new data in an index, which can be 
         index => 'test-*'
         query => '{ "query": { "range": { "event.ingested": { "gt": ":last_value", "lt": ":present"}}}, "sort": [ { "event.ingested": {"order": "asc", "format": "strict_date_optional_time_nanos", "numeric_type" : "date_nanos" } } ] }'
         tracking_field => "[event][ingested]"
-        # set a seed value to a value known to be older than any value of `event.ingested`
-        tracking_field_seed => "1980-01-01T23:59:59.999999999Z"
-        slices => 5 # optional use of slices to speed data processing, should be less than number of primary shards
+        slices => 5 # optional use of slices to speed data processing, should be equal to or less than number of primary shards
         schedule => '* * * * *' # every minute
         schedule_overlap => false # don't accumulate jobs if one takes longer than 1 minute
       }
@@ -216,7 +217,7 @@ A good use case for this feature is to track new data in an index, which can be 
 With this setup, as new documents are indexed an `test-*` index, the next scheduled run will:
 
 . select all new documents since the last observed value of the tracking field;
-. use PIT+search_after to paginate through all the data;
+. use <<point-in-time-api,Point in time (PIT)>> + <<search-after, Search after>> to paginate through all the data;
 . update the value of the field at the end of the pagination.
 
 [id="plugins-{type}s-{plugin}-options"]
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -547,7 +547,7 @@ exactly once.
   * Default value is `true`
 
 Whether to allow queuing of a scheduled run if a run is occurring.
-While this is ideal for ensuring a new run happens immediatelly after the previous on finishes if there
+While this is ideal for ensuring a new run happens immediately after the previous on finishes if there
 is a lot of work to do, but given the queue is unbounded it may lead to an out of memory over long periods of time
 if the queue grows continuously.
 
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -56,9 +56,9 @@ It is sometimes desirable to track the value of a particular field between two j
 * avoid re-processing the entire result set of a long query after an unplanned restart
 * only grab new data from an index instead of processing the entire set on each job
 
-For this, the Elasticsearch input plugin provides the <<tracking_field>> and <<tracking_field_seed>> options.
-When <<tracking_field>> is set, the plugin will record the value of that field for the last document retrieved in a run into
-a file (location defaults to <<last_run_metadata_path>>.
+For this, the Elasticsearch input plugin provides the <<plugins-{type}s-{plugin}-tracking_field>> and <<plugins-{type}s-{plugin}-tracking_field_seed>> options.
+When <<plugins-{type}s-{plugin}-tracking_field>> is set, the plugin will record the value of that field for the last document retrieved in a run into
+a file (location defaults to <<plugins-{type}s-{plugin}-last_run_metadata_path>>).
 
 The user can then inject this value in the query using the placeholder `:last_value`. The value will be injected into the query
 before execution, and the updated after the query completes, assuming new data was found.
@@ -771,8 +771,8 @@ It is also possible to target an entry in the event's metadata, which will be av
 
 Which field from the last event of a previous run will be used a cursor value for the following run.
 The value of this field is injected into each query if the query uses the placeholder `:last_value`.
-For the first query after a pipeline is started, the value used is either read from <<last_run_metadata_path>> file,
-or taken from <<tracking_field_seed>> setting.
+For the first query after a pipeline is started, the value used is either read from <<plugins-{type}s-{plugin}-last_run_metadata_path>> file,
+or taken from <<plugins-{type}s-{plugin}-tracking_field_seed>> setting.
 
 Note: The tracking value is updated only after the PIT+search_after run completes, it won't update during the search_after pagination. This is to allow use of slices.
 
@@ -782,7 +782,7 @@ Note: The tracking value is updated only after the PIT+search_after run complete
 * Value type is <<string,string>>
 * There is no default value for this setting.
 
-The starting value for the <<tracking_field>> if there is no <<last_run_metadata_path>> already.
+The starting value for the <<plugins-{type}s-{plugin}-tracking_field>> if there is no <<plugins-{type}s-{plugin}-last_run_metadata_path>> already.
 For a nanosecond timestamp based field, a suggested seed value something like `1980-01-01T23:59:59.999999999Z`.
 
 [id="plugins-{type}s-{plugin}-user"]
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -48,9 +48,65 @@ This would create an Elasticsearch query with the following format:
       "sort": [ "_doc" ]
     }'
 
+[id="plugins-{type}s-{plugin}-scheduling"]
+==== Scheduling
+
+Input from this plugin can be scheduled to run periodically according to a specific
+schedule. This scheduling syntax is powered by https://github.com/jmettraux/rufus-scheduler[rufus-scheduler].
+The syntax is cron-like with some extensions specific to Rufus (e.g. timezone support ).
+
+Examples:
+
+|==========================================================
+| `* 5 * 1-3 *`               | will execute every minute of 5am every day of January through March.
+| `0 * * * *`                 | will execute on the 0th minute of every hour every day.
+| `0 6 * * * America/Chicago` | will execute at 6:00am (UTC/GMT -5) every day.
+|==========================================================
+
+
+Further documentation describing this syntax can be found
+https://github.com/jmettraux/rufus-scheduler#parsing-cronlines-and-time-strings[here].
+
+
+[id="plugins-{type}s-{plugin}-auth"]
+==== Authentication
+
+Authentication to a secure Elasticsearch cluster is possible using _one_ of the following options:
+
+* <<plugins-{type}s-{plugin}-user>> AND <<plugins-{type}s-{plugin}-password>>
+* <<plugins-{type}s-{plugin}-cloud_auth>>
+* <<plugins-{type}s-{plugin}-api_key>>
+
+[id="plugins-{type}s-{plugin}-autz"]
+==== Authorization
+
+Authorization to a secure Elasticsearch cluster requires `read` permission at index level and `monitoring` permissions at cluster level.
+The `monitoring` permission at cluster level is necessary to perform periodic connectivity checks.
+
+[id="plugins-{type}s-{plugin}-ecs"]
+==== Compatibility with the Elastic Common Schema (ECS)
+
+When ECS compatibility is disabled, `docinfo_target` uses the `"@metadata"` field as a default, with ECS enabled the plugin
+uses a naming convention `"[@metadata][input][elasticsearch]"` as a default target for placing document information.
+
+The plugin logs a warning when ECS is enabled and `target` isn't set.
+
+TIP: Set the `target` option to avoid potential schema conflicts.
+
+[id="plugins-{type}s-{plugin}-failure-handling"]
+==== Failure handling
+
+When this input plugin cannot create a structured `Event` from a hit result, it will instead create an `Event` that is tagged with `_elasticsearch_input_failure` whose `[event][original]` is a JSON-encoded string representation of the entire hit.
+
+Common causes are:
+
+ - When the hit result contains top-level fields that are {logstash-ref}/processing.html#reserved-fields[reserved in Logstash] but do not have the expected shape. Use the <<plugins-{type}s-{plugin}-target>> directive to avoid conflicts with the top-level namespace.
+ - When <<plugins-{type}s-{plugin}-docinfo>> is enabled and the docinfo fields cannot be merged into the hit result. Combine <<plugins-{type}s-{plugin}-target>> and <<plugins-{type}s-{plugin}-docinfo_target>> to avoid conflict.
 
 [id="plugins-{type}s-{plugin}-cursor"]
 ==== Tracking a field's value across runs
+
+NOTE: experimental:[] `tracking_field` and related settings are experimental and subject to change in the future
 
 It is sometimes desirable to track the value of a particular field between two jobs:
 * avoid re-processing the entire result set of a long query after an unplanned restart
@@ -162,61 +218,6 @@ With this setup, as new documents are indexed an `test-*` index, the next schedu
 . select all new documents since the last observed value of the tracking field;
 . use PIT+search_after to paginate through all the data;
 . update the value of the field at the end of the pagination.
-
-[id="plugins-{type}s-{plugin}-scheduling"]
-==== Scheduling
-
-Input from this plugin can be scheduled to run periodically according to a specific
-schedule. This scheduling syntax is powered by https://github.com/jmettraux/rufus-scheduler[rufus-scheduler].
-The syntax is cron-like with some extensions specific to Rufus (e.g. timezone support ).
-
-Examples:
-
-|==========================================================
-| `* 5 * 1-3 *`               | will execute every minute of 5am every day of January through March.
-| `0 * * * *`                 | will execute on the 0th minute of every hour every day.
-| `0 6 * * * America/Chicago` | will execute at 6:00am (UTC/GMT -5) every day.
-|==========================================================
-
-
-Further documentation describing this syntax can be found
-https://github.com/jmettraux/rufus-scheduler#parsing-cronlines-and-time-strings[here].
-
-
-[id="plugins-{type}s-{plugin}-auth"]
-==== Authentication
-
-Authentication to a secure Elasticsearch cluster is possible using _one_ of the following options:
-
-* <<plugins-{type}s-{plugin}-user>> AND <<plugins-{type}s-{plugin}-password>>
-* <<plugins-{type}s-{plugin}-cloud_auth>>
-* <<plugins-{type}s-{plugin}-api_key>>
-
-[id="plugins-{type}s-{plugin}-autz"]
-==== Authorization
-
-Authorization to a secure Elasticsearch cluster requires `read` permission at index level and `monitoring` permissions at cluster level.
-The `monitoring` permission at cluster level is necessary to perform periodic connectivity checks.
-
-[id="plugins-{type}s-{plugin}-ecs"]
-==== Compatibility with the Elastic Common Schema (ECS)
-
-When ECS compatibility is disabled, `docinfo_target` uses the `"@metadata"` field as a default, with ECS enabled the plugin
-uses a naming convention `"[@metadata][input][elasticsearch]"` as a default target for placing document information.
-
-The plugin logs a warning when ECS is enabled and `target` isn't set.
-
-TIP: Set the `target` option to avoid potential schema conflicts.
-
-[id="plugins-{type}s-{plugin}-failure-handling"]
-==== Failure handling
-
-When this input plugin cannot create a structured `Event` from a hit result, it will instead create an `Event` that is tagged with `_elasticsearch_input_failure` whose `[event][original]` is a JSON-encoded string representation of the entire hit.
-
-Common causes are:
-
- - When the hit result contains top-level fields that are {logstash-ref}/processing.html#reserved-fields[reserved in Logstash] but do not have the expected shape. Use the <<plugins-{type}s-{plugin}-target>> directive to avoid conflicts with the top-level namespace.
- - When <<plugins-{type}s-{plugin}-docinfo>> is enabled and the docinfo fields cannot be merged into the hit result. Combine <<plugins-{type}s-{plugin}-target>> and <<plugins-{type}s-{plugin}-docinfo_target>> to avoid conflict.
 
 [id="plugins-{type}s-{plugin}-options"]
 ==== Elasticsearch Input configuration options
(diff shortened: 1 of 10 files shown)


lib/logstash/inputs/elasticsearch.rb
```diff
@@ -252,6 +269,10 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
   # exactly once.
   config :schedule, :validate => :string
 
+  # Allow scheduled runs to overlap (enabled by default). Setting to false will
+  # only start a new scheduled run after the previous one completes.
+  config :schedule_overlap, :validate => :string
```
JAndritsch, line 274: Should this be validated as a `:bool`?

JAndritsch:
I haven't had a chance to test this extensively yet, but after looking at the code and deploying the plugin for a few brief tests, I have some thoughts/observations.

Before I get to that, I'd like to provide some context as to how I'm achieving similar behavior that ensures "at least once" delivery of documents without using a cursor or sincedb-like mechanism. This will be helpful for comparing that solution to the one in this PR.

## Context

My organization uses Logstash to ship logs and metrics from one Elastic Stack to another. We handle this by using the Elasticsearch input to periodically fetch documents from relevant indices and forward them to another destination over a variety of outputs. 

To avoid resending the same docs over and over, we add a tracking field (a simple boolean type) to each document that passes through the pipeline, then save the modified document back to Elasticsearch once the pipeline's other outputs succeed. The query powering the Elasticsearch input has criteria that excludes documents where that tracking field exists. 

This has worked pretty reliably so far. It also ensures retries, as documents that were not sent successfully will be picked up on subsequent iterations of the scheduled input. However, it means we have to perform lots of writes back to the index. This is not ideal.

## Observations

Here are some of my initial thoughts after looking at this PR. I may come back and edit this with additional findings/thoughts so that they're consolidated in one comment rather than spread out across several.

### Picking the right tracking field

* All documents in the queried indices need to have the tracking field present in its source for this to work. This means there needs to be a process at ingestion time that ensures a non-blank tracking field exists for every document. This probably needs to be stated verbosely somewhere in the documentation for this plugin.
* `@timestamp` may not be reliable if historical documents are indexed with a value that occurs before the tracking field's most recently recorded value. Using `event.ingested` is probably the better choice.
* Tracking only works with time-parsable fields, which makes sense.

### Delivery guarantees and edge cases

* There's no way to recover a missed document once tracking position has moved beyond it. You can maybe try modifying the cached last value by hand, but that would be a hacky workaround. This would also mean the cursor will pick up duplicate documents on its next tick.
* Cursor tracking is updated once the event is added to the output queue, and then the input assumes it'll be handled from there and shouldn't find that same event in future iterations. If the event fails at the output, future invocations of the query will not pick it up again as tracking state was already recorded by this point. This means the output needs to be properly configured for retries, persistence, and/or a DLQ. That's something that could probably be included in the documentation for this plugin as just a reminder to those wanting to use it.
  * Alternatively, it might be interesting if we could control when the tracking data is updated. I think that would require a new output plugin to go along with this one, where users need to explicitly invoke a plugin at the output stage that updates the tracking state only after all of the pipeline's other outputs have succeeded. This would help ensure a failure to output a batch of events results in those events being picked up on the next scheduled query.
* Should `:schedule_overlap` be forced to `true` when using a tracking field? Otherwise I'm not sure what behavior we'll get if we allow schedules to overlap and change the cursor value.
* What happens if I change the tracking field in my input after a previous one has been in use? For example, maybe I start with `event.ingested` for the `logs-*` index but then later decide I'd rather use some custom timestamp field I'm adding. I can reset state by configuring the input to use a new file path for tracking, but then I won't have a way to ensure documents previously sent based on `event.ingested` won't be re-queried. I'd have to somehow include that last tracking field's value in my query to exclude those documents from being found again.

### Errors

* I'm not sure how I encountered this, but it appears the `inject_cursor` method is throwing an error for a pipeline of mine. I've configured the `tracking_field` to be `@timestamp`. When the query runs, here's a snippet of the stack trace I see:
```json
{
    "exception": "TypeError",
    "message": "no implicit conversion of LogStash::Timestamp into String",
    "backtrace": [
        "org/jruby/RubyString.java:3219:in `gsub'",
        "/usr/share/logstash/vendor/local_gems/0bcf64c0/logstash-input-elasticsearch-5.0.1/lib/logstash/inputs/elasticsearch/cursor_tracker.rb:52:in `inject_cursor'",
        "/usr/share/logstash/vendor/local_gems/0bcf64c0/logstash-input-elasticsearch-5.0.1/lib/logstash/inputs/elasticsearch.rb:375:in `get_query_object'",
        "/usr/share/logstash/vendor/local_gems/0bcf64c0/logstash-input-elasticsearch-5.0.1/lib/logstash/inputs/elasticsearch.rb:364:in `block in run'",
        "/usr/share/logstash/vendor/bundle/jruby/3.1.0/gems/rufus-scheduler-3.9.1/lib/rufus/scheduler/jobs_core.rb:241:in `block in do_call'",
        "/usr/share/logstash/vendor/bundle/jruby/3.1.0/gems/rufus-scheduler-3.9.1/lib/rufus/scheduler.rb:130:in `around_trigger'",
        "/usr/share/logstash/vendor/bundle/jruby/3.1.0/gems/rufus-scheduler-3.9.1/lib/rufus/scheduler/jobs_core.rb:240:in `do_call'",
        "/usr/share/logstash/vendor/bundle/jruby/3.1.0/gems/rufus-scheduler-3.9.1/lib/rufus/scheduler/jobs_core.rb:288:in `trigger_now'",
        "/usr/share/logstash/vendor/bundle/jruby/3.1.0/gems/rufus-scheduler-3.9.1/lib/rufus/scheduler/jobs_core.rb:339:in `block in start_work_thread'",
        "/usr/share/logstash/vendor/bundle/jruby/3.1.0/gems/rufus-scheduler-3.9.1/lib/rufus/scheduler/jobs_core.rb:342:in `block in start_work_thread'",
        "org/jruby/RubyKernel.java:1722:in `loop'",
        "/usr/share/logstash/vendor/bundle/jruby/3.1.0/gems/rufus-scheduler-3.9.1/lib/rufus/scheduler/jobs_core.rb:325:in `block in start_work_thread'"
    ]
}
```

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -126,6 +127,22 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
...
+  #       * checkpointing is only safe to do after each run (not per document)
+  config :tracking_field, :validate => :string
+
+  # Define the initial seed value of the tracking_field
+  config :tracking_field_seed, :validate => :string
+
+  # The location of where the tracking field value will be stored
+  # The value is persisted after each scheduled run (and not per result)
+  # If it's not set it defaults to '${path.data}/plugins/inputs/elasticsearch/last_run_value'
+  config :last_run_metadata_path, :validate => :string
```
JAndritsch, line 144: Might be good to call out in documentation that this needs to be configured explicitly with a unique path if you use more than one `elasticsearch` input within your pipeline, otherwise they'll share the same tracking file.

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -126,6 +127,22 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
   # by this pipeline input.
   config :slices, :validate => :number
 
+  # Enable tracking the value of a given field to be used as a cursor
+  # TODO: main concerns
+  #       * schedule overlap needs to be disabled (hardcoded as enabled)
+  #       * using anything other than _event.timestamp easily leads to data loss
+  #       * the first "synchronization run can take a long time"
+  #       * checkpointing is only safe to do after each run (not per document)
+  config :tracking_field, :validate => :string
```
JAndritsch, line 136: Since this setting only works with Time fields, is there a way to validate the mappings of the configured field?

lib/logstash/inputs/elasticsearch/cursor_tracker.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+
+    def record_last_value(event)
+      value = event.get(@tracking_field)
+      logger.trace? && logger.trace("storing last_value if #{@tracking_field} for #{Thread.current.object_id}: #{value}")
+      @last_value_hashmap.put(Thread.current.object_id, value)
+    end
+
+    def inject_cursor(query_json)
+      query_json.gsub(":last_value", @last_value)
```
JAndritsch, line 52:
```suggestion
      query_json.gsub(":last_value", @last_value.to_s)
```

JAndritsch: Fixes an implicit conversion error that results in the following trace:

```json
{
    "exception": "TypeError",
    "message": "no implicit conversion of LogStash::Timestamp into String",
    "backtrace": [
        "org/jruby/RubyString.java:3219:in `gsub'",
        "/usr/share/logstash/vendor/local_gems/0bcf64c0/logstash-input-elasticsearch-5.0.1/lib/logstash/inputs/elasticsearch/cursor_tracker.rb:52:in `inject_cursor'",
        "/usr/share/logstash/vendor/local_gems/0bcf64c0/logstash-input-elasticsearch-5.0.1/lib/logstash/inputs/elasticsearch.rb:375:in `get_query_object'",
        "/usr/share/logstash/vendor/local_gems/0bcf64c0/logstash-input-elasticsearch-5.0.1/lib/logstash/inputs/elasticsearch.rb:364:in `block in run'",
        "/usr/share/logstash/vendor/bundle/jruby/3.1.0/gems/rufus-scheduler-3.9.1/lib/rufus/scheduler/jobs_core.rb:241:in `block in do_call'",
        "/usr/share/logstash/vendor/bundle/jruby/3.1.0/gems/rufus-scheduler-3.9.1/lib/rufus/scheduler.rb:130:in `around_trigger'",
        "/usr/share/logstash/vendor/bundle/jruby/3.1.0/gems/rufus-scheduler-3.9.1/lib/rufus/scheduler/jobs_core.rb:240:in `do_call'",
        "/usr/share/logstash/vendor/bundle/jruby/3.1.0/gems/rufus-scheduler-3.9.1/lib/rufus/scheduler/jobs_core.rb:288:in `trigger_now'",
        "/usr/share/logstash/vendor/bundle/jruby/3.1.0/gems/rufus-scheduler-3.9.1/lib/rufus/scheduler/jobs_core.rb:339:in `block in start_work_thread'",
        "/usr/share/logstash/vendor/bundle/jruby/3.1.0/gems/rufus-scheduler-3.9.1/lib/rufus/scheduler/jobs_core.rb:342:in `block in start_work_thread'",
        "org/jruby/RubyKernel.java:1722:in `loop'",
        "/usr/share/logstash/vendor/bundle/jruby/3.1.0/gems/rufus-scheduler-3.9.1/lib/rufus/scheduler/jobs_core.rb:325:in `block in start_work_thread'"
    ]
}
```

strawgate:
Wow! Thanks for taking the time to provide some great feedback, we really appreciate it

> ### Picking the right tracking field
> * All documents in the queried indices need to have the tracking field present in its source for this to work. This means there needs to be a process at ingestion time that ensures a non-blank tracking field exists for every document. This probably needs to be stated verbosely somewhere in the documentation for this plugin.
> * Any documents that lack the tracking field at the time this plugin is used will either need to be modified to include that field or somehow excluded from the query by other means. The plugin supports a `tracking_field_seed` which can help account for starting at a later point in the stream, but if there are documents which have already been processed by some forwarder pipeline and do not have this tracking field set then you'll need a separate thing to exclude them.

So with a data stream, every document must have a `@timestamp` field. So we always know there will be a field we can cursor against. But we don't want to recommend you use `@timestamp` for the obvious reasons. So, we'd tell you to use `event.ingested` as the field instead. And if you're using data streams without our integrations, that means you'd need to ensure `event.ingested` gets added to your documents.

But once you have done this, likely via an ingest pipeline, grabbing the documents from before that point (that do not have event.ingested) is actually pretty simple. With all new documents being stamped with `event.ingested`, you can now use one elasticsearch input to grab all documents that have `event.ingested` (to send all new data) and you can use another input targeting`@timestamp` to grab all documents that don't have the `event.ingested` field (to backfill historical data):
```
input {
  elasticsearch {
    id => eventingestedcursor
    query => '{ "query": { "range": {"event.ingested": { "gt": ":last_value", "lte": "now"}}}, "sort": [ { "event.ingested": "asc" } ] }'
    tracking_field => "[event][ingested]"
    slices => 2
    schedule => '*/30 * * * * *'
  }
  elasticsearch {
    id => timestampcursor
    query => '{ "query": {"query":{"bool":{"must_not":[{"exists":{"field":"event.ingested"}}],"must":[],"filter":[{"range":{"@timestamp":{"gte":":last_value"}}}],"should":[]}}}, "sort": [ { "@timestamp": "asc" } ] }'
    tracking_field => "@timestamp"
    slices => 2
    schedule => '*/30 * * * * *'
  }
}
```

So now let's imagine you've built a fantastic alternative to the cursor (like your `reporting.transferred: true`), but you want to switch to our input cursors. You want to exclude any documents that don't have `reporting.transferred`. Well, you simply add the `event.ingested` ingest pipeline and add `not exists reporting.transferred` to both queries:
1. Adding it to the `@timestamp` input ensuring that any documents that got ingested between the time where the `reporting.transferred: true` step is removed and the `event.ingested` is added don't get left behind.
2. Adding it to the `event.ingested` input ensures that any documents that got ingested after `event.ingested` was added but before the `reporting.transferred: true` step was removed do not get sent twice.

> ### Delivery guarantees and edge cases
> * There's no way to recover a missed document once tracking position has moved beyond it. You can maybe try modifying the cached last value by hand, but that would be a hacky workaround. This would also mean the cursor will pick up duplicate documents on its next tick.
>   * This is in contrast to my current solution, which accounts for this by saving tracking status on each document. In this scenario, it doesn't matter when the event occurred, time-wise. The absence of the tracking field means the query will find it on next run.
> * Cursor tracking is updated once the event is added to the output queue, and then the input assumes it'll be handled from there and shouldn't find that same event in future iterations. If the event fails at the output, future invocations of the query will not pick it up again as tracking state was already recorded by this point. This means the output needs to be properly configured for retries, persistence, and/or a DLQ. That's something that could probably be included in the documentation for this plugin as just a reminder to those wanting to use it.

This is actually also the case when using a boolean field on the documents. Because you need to use two outputs (1 to rewrite the document and 1 to send data to the remote destination), and that is not atomic (it's not even ordered) -- if the write to your destination fails, your write to the local system will almost certainly succeed. In this case you have marked the documents as forwarded while actual forwarding has failed. If you're using a memory queue and Logstash restarts while the remote output is failing, you will have documents marked as transferred that were actually still being retried by the remote output.

Ultimately, for both cases, if the destination output fails entirely, the default configuration is that the output retries indefinitely. For both cases, using a persistent queue ensures that an unexpected restart of Logstash or a force shutdown of a pipeline doesn't while the output is failing doesn't resolve in data loss.

For both cases, capturing document-level failures requires a DLQ, and in both cases the tracking mechanism will consider documents in the DLQ as having been forwarded.

But I believe there's actually a big advantage to the input cursor once you reach steady state and are no longer pulling 10s of thousands of documents per query. In this case, when the output is blocked, the input is allowed to keep pushing events into the queue until the queue fills and that then blocks the input. With the cursor, the timestamp increments for at the end of each query, until the queue fills up and then blocks, each batch pushed into the queue is new documents.

With the `reporting.transferred` solution, when the destination output blocks, the `reporting.transferred` output also blocks, and each subsequent invocation of the input will return the same documents that are already in the queue, until the queue fills up. When the output becomes unblocked, you'll find the queue is full of many copies of each document.

> * Should `:schedule_overlap` be forced to `true` when using a tracking field? Otherwise I'm not sure what behavior we'll get if we allow schedules to overlap and change the cursor value.

A great question for @jsvd for sure. My read is that either `true` or `false` should be fine with them leading to slightly different, but acceptable, behavior -- it looks like [we force the max worker threads to 1 for each instance of the scheduler](https://github.com/logstash-plugins/logstash-mixin-scheduler/blob/main/lib/logstash/plugin_mixins/scheduler/rufus_impl.rb#L25C13-L25C29). So the scheduler will only ever run 1 thread at a time. [When overlap is `true`, jobs will continue to schedule into the work queue](https://github.com/jmettraux/rufus-scheduler/blob/ec048146d12d2055566ca3aff48f31d14b075fa3/lib/rufus/scheduler/jobs_core.rb#L255-L273) when the worker thread is running. When `overlap` is false, jobs will not schedule into the work queue while the work thread is running. I think the only real impact this has is that if your cron job is set to every 10 minutes, and the job takes 15 minutes, when you have `overlap: false`, it will take 5 more minutes before the next job runs, with `overlap:true` the next job will run immediately.

> * What happens if I change the tracking field in my input after a previous one has been in use? For example, maybe I start with `event.ingested` for the `logs-*` index but then later decide I'd rather use some custom timestamp field I'm adding. I can reset state by configuring the input to use a new file path for tracking, but then I won't have a way to ensure documents previously sent based on `event.ingested` won't be re-queried. I'd have to somehow include that last tracking field's value in my query to exclude those documents from being found again.

You'd setup a third input, no need to reset state:
1. Reads timestamp if (event.ingested does not exist and event.new_ingested does not exist)
2. Reads event.ingested (if event.new_ingested does not exist)
3. Reads event.new_ingested

strawgate:
Some notes on query recommendations:
The [PIT docs recommend](https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html#search-after) `"strict_date_optional_time_nanos", "numeric_type" : "date_nanos" ` on the sort to ensure that sorting treats `date` and `date_nanos` the same.

When the user wants to backfill data, and decides to set `tracking_field_seed`, we can also perhaps recommend using `date math` in the query to limit the size of the scroll and to limit long-running jobs in the input. I believe the syntax would be something like: `{ "range": {"event.ingested": { "gt": ":last_value", "lte": ":last_value||+1d"}}}` for grabbing at most 1 day of data at a time. An input set to scroll 1 day of data with each run, and having a new run scheduled every 4 minutes, would be able to backfill up to a year of data per day.

JAndritsch:
Hey @strawgate, thanks for jumping in with a super detailed response!

> I think about this as two things: backfilling historical data and streaming new data. Both are solved with the cursor but in different ways...

Backfill versus Streaming is a good distinction to point out. I had been thinking about how I might solve these problems with a single input, and using two of them with different query criteria hadn't crossed my mind. But that does seem like the right way to go about it. 

> So now let's imagine you've built a fantastic alternative to the cursor (like your reporting.transferred: true), but you want to switch to our input cursors. You want to exclude any documents that don't have reporting.transferred. Well, you simply add the event.ingested ingest pipeline and add not exists reporting.transferred to both queries...

Yeah, that makes a lot of sense and I believe addresses a hypothetical scenario where someone using this cursor tracking feature starts out with `event.ingested` but then later decides to move to a different field for tracking. They'll need to maintain a backwards-compatible input that knows to exclude pulling documents that may overlap between the new input.

Quoting myself here...

> Should :schedule_overlap be forced to true when using a tracking field? Otherwise I'm not sure what behavior we'll get if we allow schedules to overlap and change the cursor value.

I meant forced to `false`. Must not have had enough coffee when I initially typed the question XD.

Maybe it doesn't matter given the outcome isn't really that impactful as you had suggested.

> This is actually also the case when using a boolean field on the documents. Because you need to use two outputs (1 to rewrite the document and 1 to send data to the remote destination), and that is not atomic (it's not even ordered) -- if the write to your destination fails, your write to the local system will almost certainly succeed. In this case, you have marked the documents as forwarded while actual forwarding has failed. If you're using a memory queue and Logstash restarts while the remote output is failing, you will have documents marked as transferred that were actually still being retried by the remote output.

> Ultimately, for both cases, if the destination output fails entirely, the default configuration is that the output retries 
indefinitely. For both cases, using a persistent queue ensures that an unexpected restart of Logstash or a force shutdown of a pipeline while the output is failing doesn't result in data loss.

> For both cases, capturing document-level failures requires a DLQ, and in both cases the tracking mechanism will consider documents in the DLQ as having been forwarded.

> But I believe there's actually a big advantage to the input cursor once you reach steady state and are no longer pulling 10s of thousands of documents per query. In this case, when the output is blocked, the input is allowed to keep pushing events into the queue until the queue fills and that then blocks the input. With the cursor, the timestamp increments at the end of each query, until the queue fills up and then blocks, each batch pushed into the queue is new documents.

> With the reporting.transferred solution, when the destination output blocks, the reporting.transferred output also blocks. With the reporting.transferred output blocked, each subsequent invocation of the ES query will return the same documents that are already in the queue, pushing the same documents into the queue over and over again until it fills up.

I've been mulling over your feedback and thinking about how cursor tracking would work in comparison to what I have now in a few scenarios. I think most of my concerns around using a cursor would be addressed by switching from an in-memory queue to a persistent queue + DLQ.

Using a tracking field on a document in Elasticsearch sort of emulates the idea of a persistent queue state, where under certain circumstances like a pipeline reload or Logstash restart, the document doesn't get marked as delivered and thereby allows it to be re-queried on the next input tick. But if I switched to a persistent queue then that gives me the same thing, without the added cost of having to write state back to documents in bulk. And the fact that the input wouldn't be able to queue up duplicate events that are currently in the queue or being worked on is a huge win for many reasons.

I'm going to follow your suggestion of setting up multiple inputs to handle the backfill versus streaming scenarios and post an update once I've had a chance to see it work.

Thanks again!

strawgate:
@JAndritsch I shared a tool with you where you can setup various scenarios and measure data loss. I added your scenarios so you can play around with it. It's actually quite difficult to lose data with the cursor + memory queue for the backfill scenario as currently the whole read is 1 scroll, and it only updates the cursor after the whole read. I think you'll be surprised by how much data is lost when using the `reporting.transferred` method + a MQ

lib/logstash/inputs/elasticsearch.rb
jsvd: Indeed, we can harden this by opening the file in exclusive mode + document the constraint.

lib/logstash/inputs/elasticsearch.rb
jsvd: Good question. there's a major issue with this feature in that we need the user to set up things in just the right way:

1. use a pipeline that adds `_ingest.timestamp` to a field like `event.ingested_at`
  a) knowing this was set would be difficult as the user may add this field either in a dedicated or existing ingest pipeline
  b) when starting, the destination index may not have a mapping for the field as the index or field may not exist yet. Even a mapping may not exist yet as the user may be relying on dynamic mapping
2. set that field in the `tracking_field` using logstash's field reference notation (`[event][ingested_at]` vs `event.ingested_at`)
  a) we can't know this field matches the one you configured in the ingest pipeline, given 1 a) and 1 b)
3. define a range query in the right way and with a very precise sorting incantation like `"sort": [{"event.ingested_at": {"order": "asc", "format": "strict_date_optional_time_nanos", "numeric_type" : "date_nanos" }}]`
  a) allowing the user to configure any query makes it hard to verify it's the right one, it'd require a lot of regexing
  b) the alternative is to provide this feature as a black box, but would impede use cases we discussed in this issue around custom queries for "catching up to old data" using use case specific value markers.

JAndritsch:
@strawgate Finally got a chance to play with the tool and added some new scenarios.

You were right -- it's very surprising to see the cursor results in better data reliability even without a persistent queue. There's more I need to dig into here to better understand why.

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -330,25 +351,38 @@ def register
...
+
     @client
   end
 
   def run(output_queue)
     if @schedule
-      scheduler.cron(@schedule) { @query_executor.do_run(output_queue) }
+      scheduler.cron(@schedule, :overlap => @schedule_overlap) do
+        @query_executor.do_run(output_queue, get_query_object())
+        @cursor_tracker.checkpoint_cursor
```
JAndritsch, line 363: @jsvd So I may have misunderstood how checkpointing and last value tracking works on my initial pass. It looks like there are two phases:

1. The `last_value` is captured in a concurrent hashmap in memory whenever we queue events for output.
2. At some point later, we reduce down to the highest timestamp that we've cached and write it to disk.

At what point does the second step occur in relation to the pipeline's output(s)? I'm assuming this checkpointing process still happens before the outputs have completed, so there might be a slim chance that tracking state persisted before the queue was fully drained. Is that correct?

strawgate: At the moment tracking state persists when the scroll finishes. 

With the persistent queue we checkpoint the cursor after the data lands in the persistent queue.

In my other comment in the discussion you can read my two suggestions for improving this:
1. we should checkpoint the cursor after each page in the scroll 
Or
2. We should add the features required to enable short-lived scrolls

jsvd:
I've done a few modifications:

I've introduced `:present` as a second placeholder to pin down the concept of `now-30s` in the initial PIT+search and all the subsequent search_after queries within the same PIT.
Using a range query of `:last_value` up to `:present` ensures range filter is cached during all queries within the same job run, and avoids data loss at the edge of the PIT timestamps.

This significantly improves the tailing of an index use case, I was able to consistently tail an index being written to, using slices, without any data loss.

For catch-up use cases where it is known data already exists in the index, a range such as `:last_value` to `:last_value||1w` can be used instead, combined with a second range filter to query up to a known boundary (e.g. up to January 31st).

JAndritsch:
Thanks, @jsvd! I'll find a bit of time this week to test out the latest updates and report back on how that goes.

docs/index.asciidoc
```diff
@@ -622,6 +763,27 @@ When the `target` is set to a field reference, the `_source` of the hit is place
...
+
+* Value type is <<string,string>>
+* There is no default value for this setting.
+
+Which field from the last event of a previous run will be used a cursor value for the following run.
+The value of this field is injected into each query if the query uses the placeholder `:last_value`.
+For the first query after a pipeline is started, the value used is either read from <<plugins-{type}s-{plugin}-last_run_metadata_path>> file,
+or taken from <<plugins-{type}s-{plugin}-tracking_field_seed>> setting.
+
+Note: The tracking value is updated only after the PIT+search_after run completes, it won't update during the search_after pagination. This is to allow use of slices.
```
jsvd, line 777:
```suggestion
Note: The tracking value is updated after each page is read and at the end of each Point in Time. In case of a crash the last saved value will be used so some duplication of data can occur. For this reason the use of unique document IDs for each event is recommended in the downstream destination.
```

jsvd:
A few more modifications are in place: now the cursor is persisted per search + per slice and at the end of the PIT.

So slices is set to 5 and there's a PIT open followed by 10 queries per slice, there will be 50 checkpoints and a final one when closing the PIT.

This means, in a backfilling use case, termination of Logstash will not require processing of the entire original date range. 

Also we should note that in this example the first 50 checkpoints the persisted value corresponds to the earliest value between all slices, and in the last checkpoint it is the latest value that is persisted.

This ensures that if Logstash is restarted between PITs there won't be duplicate events, but that a minimal amount of duplicates can happen if the termination occurs while the PIT is open and queries are being executed.

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -124,6 +125,21 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
   # by this pipeline input.
   config :slices, :validate => :number
 
+  # Enable tracking the value of a given field to be used as a cursor
+  # TODO: main concerns
+  #       * using anything other than _event.timestamp easily leads to data loss
+  #       * the first "synchronization run can take a long time"
+  #       * checkpointing is only safe to do after each run (not per document)
```
jsvd, line 132:
```suggestion
```

lib/logstash/inputs/elasticsearch.rb
jsvd: I've changed the plugin to store the file in a folder namespaced by the pipeline id:

```asciidoc
[id="plugins-{type}s-{plugin}-last_run_metadata_path"]
===== `last_run_metadata_path`

  * Value type is <<string,string>>
  * There is no default value for this setting.

The path to store the last observed value of the tracking field, when used.
By default this file is stored as `<path.data>/plugins/inputs/elasticsearch/<pipeline_id>/last_run_value`.

This setting should point to file, not a directory, and Logstash must have read+write access to this file.
```

docs/index.asciidoc
```diff
@@ -622,6 +765,27 @@ When the `target` is set to a field reference, the `_source` of the hit is place
...
+For the first query after a pipeline is started, the value used is either read from <<plugins-{type}s-{plugin}-last_run_metadata_path>> file,
+or taken from <<plugins-{type}s-{plugin}-tracking_field_seed>> setting.
+
+Note: The tracking value is updated after each page is read and at the end of each Point in Time. In case of a crash the last saved value will be used so some duplication of data can occur. For this reason the use of unique document IDs for each event is recommended in the downstream destination.
+
+[id="plugins-{type}s-{plugin}-tracking_field_seed"]
+===== `tracking_field_seed`
+
+* Value type is <<string,string>>
+* There is no default value for this setting.
```
robbavey, line 785: Looking at the code it defaults to `Time.now`?

docs/index.asciidoc
```diff
@@ -622,6 +765,27 @@ When the `target` is set to a field reference, the `_source` of the hit is place
...
+
+The starting value for the <<plugins-{type}s-{plugin}-tracking_field>> if there is no <<plugins-{type}s-{plugin}-last_run_metadata_path>> already.
+For a nanosecond timestamp based field, a suggested seed value something like `1980-01-01T23:59:59.999999999Z`.
```
robbavey, line 788: Maybe an explanation of behavior when a `tracking_field` is chosen without a `tracking_field_seed`?

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -640,6 +678,29 @@ def setup_query_executor
...
 
+  def setup_cursor_tracker
+    return unless @tracking_field
+    return unless @query_executor.is_a?(LogStash::Inputs::Elasticsearch::SearchAfter)
+
+    if @resolved_search_api != "search_after" || @response_type != "hits"
+      raise ConfigurationError.new("The `tracking_field` feature can only be used with `search_after` non-aggregation queries")
+    end
+
+    @tracking_field_seed ||= Time.now.utc.iso8601
```
robbavey, line 689: Would it make sense to default to, say, epoch, to default to catching as many messages as possible, rather than defaulting to skipping past events ingested prior to the first time this plugin is run?

lib/logstash/inputs/elasticsearch/cursor_tracker.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    def inject_cursor(query_json)
+      # ":present" means "now - 30s" to avoid grabbing partially visible data in the PIT
+      result = query_json.gsub(":last_value", @last_value.to_s).gsub(":present", now_in_nanos)
+      logger.debug("inject_cursor: injected values for ':last_value' and ':present'", :query => result)
+      result
+    end
+
+    def now_in_nanos
```
robbavey, line 54: Maybe this could be `thirty_seconds_ago_in_nanos`, to avoid confusion, or `seconds_ago_in_nanos(num_of_seconds)` to avoid confusion

docs/index.asciidoc
```diff
@@ -410,6 +540,19 @@ for example: "* * * * *" (execute query every minute, on the minute)
...
 exactly once.
 
+[id="plugins-{type}s-{plugin}-schedule_overlap"]
+===== `schedule_overlap`
+
+  * Value type is <<boolean,boolean>>
+  * Default value is `true`
+
+Whether to allow queuing of a scheduled run if a run is occurring.
+While this is ideal for ensuring a new run happens immediatelly after the previous on finishes if there
```
robbavey, line 550:
```suggestion
While this is ideal for ensuring a new run happens immediately after the previous on finishes if there
```

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -250,6 +265,10 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
   # exactly once.
   config :schedule, :validate => :string
 
+  # Allow scheduled runs to overlap (enabled by default). Setting to false will
+  # only start a new scheduled run after the previous one completes.
+  config :schedule_overlap, :validate => :boolean
```
robbavey, line 270: Should we force this to `false` if the tracking cursor is in play?

robbavey (changes requested):
First pass through - great stuff

lib/logstash/inputs/elasticsearch.rb
jsvd: not necessarily, if the user knows there are periods of more intense ingestion, they may want to keep overlap, so there's not downtime between runs during hot periods, but they know outside of those there won't be job accumulation.

docs/index.asciidoc
jsvd: no need now that seed has a default to epoch.

docs/index.asciidoc
```diff
@@ -103,6 +103,122 @@ Common causes are:
...
+a file (location defaults to <<plugins-{type}s-{plugin}-last_run_metadata_path>>).
+
+The user can then inject this value in the query using the placeholder `:last_value`. The value will be injected into the query
+before execution, and the updated after the query completes, assuming new data was found.
+
+This feature works best when:
+* the query sorts by the tracking field
+* the field type has enough resolution so that two events are unlikely to have the same value for the field
+
+The plugin also offers another placeholder called `:present` used to inject the nano-second based value of "now-30s".
```
robbavey, line 126: Would it make sense to move this placeholder description below the suggestion and introduce it with something along the lines of:

"To support this, the plugin also offers a placeholder called..."

robbavey (approved):
Some minor doc suggestions, but other than that LGTM

lib/logstash/inputs/elasticsearch.rb
strawgate: I think the important detail here is that the new scheduled run doesn't actually start running until the previous one finishes, instead it is simply scheduled to start and will start immediately when the previous one finishes, right?

jsvd: correct given the the threads == 1, so only 1 job runs at a time regardless of this setting's value.
With overlap new jobs will be queued if a job is running, without overlap new jobs are dropped if a job is running.

docs/index.asciidoc
```diff
@@ -103,6 +103,123 @@ Common causes are:
  - When the hit result contains top-level fields that are {logstash-ref}/processing.html#reserved-fields[reserved in Logstash] but do not have the expected shape. Use the <<plugins-{type}s-{plugin}-target>> directive to avoid conflicts with the top-level namespace.
  - When <<plugins-{type}s-{plugin}-docinfo>> is enabled and the docinfo fields cannot be merged into the hit result. Combine <<plugins-{type}s-{plugin}-target>> and <<plugins-{type}s-{plugin}-docinfo_target>> to avoid conflict.
 
+[id="plugins-{type}s-{plugin}-cursor"]
+==== Tracking a field's value across runs
+
+NOTE: experimental:[] `tracking_field` and related settings are experimental and subject to change in the future
+
```
karenzone, line 110:
```suggestion
.Technical Preview: Tracking a field's value
****
The feature that allows tracking a field's value across runs is in _Technical Preview_.
Configuration options and implementation details are subject to change in minor releases without being preceded by deprecation warnings.
****

```

docs/index.asciidoc
```diff
@@ -103,6 +103,123 @@ Common causes are:
...
+It is sometimes desirable to track the value of a particular field between two jobs:
+* avoid re-processing the entire result set of a long query after an unplanned restart
+* only grab new data from an index instead of processing the entire set on each job
```
karenzone, line 113:
```suggestion
Some uses cases require tracking the value of a particular field between two jobs.
Examples include:

* avoiding the need to re-process the entire result set of a long query after an unplanned restart
* grabbing only new data from an index instead of processing the entire set on each job.
```

docs/index.asciidoc
```diff
@@ -103,6 +103,123 @@ Common causes are:
...
+
+For this, the Elasticsearch input plugin provides the <<plugins-{type}s-{plugin}-tracking_field>> and <<plugins-{type}s-{plugin}-tracking_field_seed>> options.
+When <<plugins-{type}s-{plugin}-tracking_field>> is set, the plugin will record the value of that field for the last document retrieved in a run into
+a file (location defaults to <<plugins-{type}s-{plugin}-last_run_metadata_path>>).
```
karenzone, line 117:
```suggestion
The Elasticsearch input plugin provides the <<plugins-{type}s-{plugin}-tracking_field>> and <<plugins-{type}s-{plugin}-tracking_field_seed>> options.
When <<plugins-{type}s-{plugin}-tracking_field>> is set, the plugin records the value of that field for the last document retrieved in a run into
a file. 
(The file location defaults to <<plugins-{type}s-{plugin}-last_run_metadata_path>>.)
```

docs/index.asciidoc
```diff
@@ -103,6 +103,123 @@ Common causes are:
...
+
+The user can then inject this value in the query using the placeholder `:last_value`. The value will be injected into the query
+before execution, and the updated after the query completes if new data was found.
```
karenzone, line 120:
```suggestion
You can then inject this value in the query using the placeholder `:last_value`. 
The value will be injected into the query before execution, and then updated after the query completes if new data was found.
```

docs/index.asciidoc
```diff
@@ -103,6 +103,123 @@ Common causes are:
...
+
+This feature works best when:
+
+. the query sorts by the tracking field;
+. the timestamp field is added by {es};
+. the field type has enough resolution so that two events are unlikely to have the same value.
```
karenzone, line 126:
```suggestion
* the query sorts by the tracking field,
* the timestamp field is added by {es}, and
* the field type has enough resolution so that two events are unlikely to have the same value.
```

docs/index.asciidoc
```diff
@@ -103,6 +103,123 @@ Common causes are:
...
+
+It is recommended to use a tracking field whose type is https://www.elastic.co/guide/en/elasticsearch/reference/current/date_nanos.html[date nanoseconds].
+If the tracking field is of this data type, an extra placeholder called `:present` can be used to inject the nano-second based value of "now-30s".
+This placeholder is useful as the right-hand side of a range filter, allowing the collection of
+new data but leaving partially-searcheable bulk request data to the next scheduled job.
+
+Below is a series of steps to help set up the "tailing" of data being written to a set of indices, using a date nanosecond field
+added by an Elasticsearch ingest pipeline, and the `tracking_field` capability of this plugin.
```
karenzone, line 134:
```suggestion
id="plugins-{type}s-{plugin}-tracking-sample"]
===== Sample configuration: Track field value across runs

This section contains a series of steps to help you set up the "tailing" of data being written to a set of indices, using a date nanosecond field added by an Elasticsearch ingest pipeline and the `tracking_field` capability of this plugin.
```

karenzone (commented):
Nice work on these docs! The config samples are especially helpful.
I left some suggestions inline for consideration, mostly around style, capitalization, indents, etc.  Please use what you like.

Ping me when you're ready for final check.  I'll update the content in this [test PR](https://github.com/elastic/logstash-docs/pull/1828) to make sure everything checks out.

docs/index.asciidoc
karenzone: We don't need both notes, so please decide which treatment and working you like.

docs/index.asciidoc
```diff
@@ -103,6 +103,134 @@ Common causes are:
  - When the hit result contains top-level fields that are {logstash-ref}/processing.html#reserved-fields[reserved in Logstash] but do not have the expected shape. Use the <<plugins-{type}s-{plugin}-target>> directive to avoid conflicts with the top-level namespace.
  - When <<plugins-{type}s-{plugin}-docinfo>> is enabled and the docinfo fields cannot be merged into the hit result. Combine <<plugins-{type}s-{plugin}-target>> and <<plugins-{type}s-{plugin}-docinfo_target>> to avoid conflict.
 
+[id="plugins-{type}s-{plugin}-cursor"]
+==== Tracking a field's value across runs
+
+NOTE: experimental:[] `tracking_field` and related settings are experimental and subject to change in the future
+.Technical Preview: Tracking a field's value
```
jsvd, line 110:
```suggestion

.Technical Preview: Tracking a field's value
```

docs/index.asciidoc
```diff
@@ -103,6 +103,133 @@ Common causes are:
...
+* the query sorts by the tracking field,
+* the timestamp field is added by {es}, and
+* the field type has enough resolution so that two events are unlikely to have the same value.
+
+Consider using a tracking field whose type is https://www.elastic.co/guide/en/elasticsearch/reference/current/date_nanos.html[date nanoseconds].
+If the tracking field is of this data type, you can use an extra placeholder called `:present` to inject the nano-second based value of "now-30s".
+This placeholder is useful as the right-hand side of a range filter, allowing the collection of
+new data but leaving partially-searchable bulk request data to the next scheduled job.
+
+id="plugins-{type}s-{plugin}-tracking-sample"]
```
karenzone, line 140:
```suggestion
[id="plugins-{type}s-{plugin}-tracking-sample"]
```

karenzone: Formatting error.  This one's on me.

karenzone (approved):
LGTM! 🚀 

Mocked up preview is here: https://logstash-docs_bk_1828.docs-preview.app.elstc.co/guide/en/logstash/8.x/plugins-inputs-elasticsearch.html

PR #207, opened 2024-07-05
Adding support for custom headers

flexitrev:
This adds support for custom headers to be specified in the input configuration and included in transport options. 

custom_headers => {"Header Name" => "Value"}

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -214,7 +214,7 @@ Connect timeouts tend to occur when Elasticsearch or an intermediate proxy is ov
 
 Pass a set of key value pairs as the headers sent in each request to an elasticsearch node.
 The headers will be used for any kind of request.
-These custom headers will be overridden by settings such as `compression_level`.
+These custom headers will override any headers previously set by the plugin such as the User Agent or Authorization headers.
 
 [id="plugins-{type}s-{plugin}-docinfo"]
 ===== `docinfo`
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -23,7 +23,7 @@ include::{include_path}/plugin_header.asciidoc[]
 
 Read from an Elasticsearch cluster, based on search query results.
 This is useful for replaying test logs, reindexing, etc.
-You can periodically schedule ingestion using a cron syntax 
+You can periodically schedule ingestion using a cron syntax
 (see `schedule` setting) or run the query one time to load
 data into Logstash.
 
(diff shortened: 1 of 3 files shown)


jsvd (changes requested):
missing doc entry in asciidoc

docs/index.asciidoc
```diff
@@ -199,6 +200,16 @@ For more info, check out the
...
 
+[id="plugins-{type}s-{plugin}-custom_headers"]
+===== `custom_headers`
+
+  * Value type is <<hash,hash>>
+  * Default value is empty
+
+Pass a set of key value pairs as the headers sent in each request to
+an elasticsearch node. The headers will be used for any kind of request
+These custom headers will be overidden by settings like `compression_level`.
```
karenzone, line 211:
```suggestion
Pass a set of key value pairs as the headers sent in each request to an elasticsearch node. 
The headers will be used for any kind of request.
These custom headers will be overridden by settings such as `compression_level`.
```

karenzone: Please use asciidoctor's "[one sentence per line](https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line)" recommendation, as it offers some pretty sweet benefits.

karenzone (commented):
Left a comment inline for your consideration.

docs/index.asciidoc
```diff
@@ -205,8 +206,18 @@ For more info, check out the
...
 
+[id="plugins-{type}s-{plugin}-custom_headers"]
+===== `custom_headers`
+
+  * Value type is <<hash,hash>>
+  * Default value is empty
+
+Pass a set of key value pairs as the headers sent in each request to an elasticsearch node.
+The headers will be used for any kind of request.
+These custom headers will be overridden by settings such as `compression_level`.
```
jsvd, line 217: This plugin doesn't have a `compression_level` setting, and it will override other headers since the setting is evaluated after others like user agent and authorization:

```suggestion
These custom headers will override any headers previously set by the plugin such as the User Agent or Authorization headers.
```

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -298,6 +301,7 @@ def register
     transport_options[:headers].merge!(setup_basic_auth(user, password))
     transport_options[:headers].merge!(setup_api_key(api_key))
     transport_options[:headers].merge!({'user-agent' => prepare_user_agent()})
+    transport_options[:headers].merge!(@custom_headers) if @custom_headers
```
jsvd, line 304:
```suggestion
    transport_options[:headers].merge!(@custom_headers) unless @custom_headers.empty?
```

jsvd (approved):
LGTM

Issue #208, opened 2024-08-15
"response_type" = aggregation ignores "index" parameter


vulnerabivoro:
Found this:

- removing the `response_type : "aggregations"` parameter , logstash uses the `PIT` API of the index set in parameters:

```
[DEBUG] 2024-08-15 11:32:44.219 [Converge PipelineAction::Create<main>] elasticsearch - config LogStash::Inputs::Elasticsearch/@index = ".internal.alerts-security.alerts-default-000026"
[DEBUG] 2024-08-15 11:33:00.088 [Ruby-0-Thread-60@[main]|input|Elasticsearch|scheduler_worker-00: /usr/share/logstash/vendor/bundle/jruby/3.1.0/gems/rufus-scheduler-3.9.1/lib/rufus/scheduler/jobs_core.rb:315] MainClientExec - Executing request POST /.internal.alerts-security.alerts-default-000026/_pit?keep_alive=1m HTTP/1.1
[DEBUG] 2024-08-15 11:33:00.088 [Ruby-0-Thread-60@[main]|input|Elasticsearch|scheduler_worker-00: /usr/share/logstash/vendor/bundle/jruby/3.1.0/gems/rufus-scheduler-3.9.1/lib/rufus/scheduler/jobs_core.rb:315] headers - http-outgoing-0 >> POST /.internal.alerts-security.alerts-default-000026/_pit?keep_alive=1m HTTP/1.1
[DEBUG] 2024-08-15 11:33:00.088 [Ruby-0-Thread-60@[main]|input|Elasticsearch|scheduler_worker-00: /usr/share/logstash/vendor/bundle/jruby/3.1.0/gems/rufus-scheduler-3.9.1/lib/rufus/scheduler/jobs_core.rb:315] wire - http-outgoing-0 >> "POST /.internal.alerts-security.alerts-default-000026/_pit?keep_alive=1m HTTP/1.1[\r][\n]"
```

- with `response_type : "aggregations"`parameter, the plugin doesnt use the pit API of the index to make the call, and just executes against `_search` API, with no specific index pointed out.

```
[...]

PR #209, opened 2024-08-22
Fix aggregation to respect `index` param

kaisecheng:
The `index` is ignored due to always pass a nil value [here](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/202/files#diff-d3674affa8241cc524249e72d865f344e543818029d840f709173e80b98ba309R19)
This commit passes the right param and adds unit test

Fixed: https://github.com/logstash-plugins/logstash-input-elasticsearch/issues/208


vulnerabivoro:
related to #https://github.com/logstash-plugins/logstash-input-elasticsearch/issues/208

lib/logstash/inputs/elasticsearch/aggregation.rb
```diff
@@ -16,7 +16,7 @@ def initialize(client, plugin)
           @query = @plugin_params["query"]
           @retries = @plugin_params["retries"]
           @agg_options = {
-            :index => @index,
+            :index => @plugin_params["index"],
             :size  => 0
```
mashhurs, line 20: Good catch @kaisecheng!
How about size?

yaauie: The `size` should still be `0`, because this is specifically an _aggregations-only_ query and we don't need query results.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 4.20.4
+  - Fix issue where the `index` parameter was being ignored during aggregation [#209](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/209)
```
yaauie, line 2:
```suggestion
  - Fix issue where the `index` parameter was being ignored when using `response_type => aggregation` [#209](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/209)
```

kaisecheng:
CI was red  because centos 7 is end of life and the mirror list has been removed, hence, docker build failed when testing aginst Logstash 7.16.3. 
remove legacy test from travis [43a9ed1](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/209/commits/43a9ed1b5b546ff63b5ede9328d18b55088a9efc)

mashhurs (approved):
lgtm!

PR #213, opened 2024-11-12
Remove deprecated SSL settings and simplify SSL configuration

donoghuc:
Get rid of the deprecated parameters and document their removal. Ensure the integration and unit tests pass and that the removals are clearly documented. 

Closes https://github.com/logstash-plugins/logstash-input-elasticsearch/issues/210

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
-## 4.20.5
-  - Add `x-elastic-product-origin` header to Elasticsearch requests [#211](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/211)
+## 5.0.0 ##
+  - SSL settings that were marked deprecated in version `4.17.0` are now marked obsolete, and will prevent the plugin from starting.
+  - These settings are:
+    - `ssl`, which should bre replaced by `ssl_enabled`
+    - `ca_file`, which should bre replaced by `ssl_certificate_authorities`
+    - `ssl_certificate_verification`, which should bre replaced by `ssl_verification_mode`
+    - [#213](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/213)
 
 ## 4.20.4
   - Fix issue where the `index` parameter was being ignored when using `response_type => aggregations` [#209](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/209)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,13 @@
-## 5.0.0 ##
+## 5.0.0
   - SSL settings that were marked deprecated in version `4.17.0` are now marked obsolete, and will prevent the plugin from starting.
   - These settings are:
     - `ssl`, which should bre replaced by `ssl_enabled`
     - `ca_file`, which should bre replaced by `ssl_certificate_authorities`
     - `ssl_certificate_verification`, which should bre replaced by `ssl_verification_mode`
     - [#213](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/213)
+
+## 4.20.5
+  - Add `x-elastic-product-origin` header to Elasticsearch requests [#211](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/211)
 
 ## 4.20.4
   - Fix issue where the `index` parameter was being ignored when using `response_type => aggregations` [#209](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/209)
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -96,7 +96,13 @@ TIP: Set the `target` option to avoid potential schema conflicts.
 [id="plugins-{type}s-{plugin}-options"]
 ==== Elasticsearch Input configuration options
 
-This plugin supports the following configuration options plus the <<plugins-{type}s-{plugin}-common-options>> and the <<plugins-{type}s-{plugin}-deprecated-options>> described later.
+This plugin supports these configuration options plus the <<plugins-{type}s-{plugin}-common-options>> described later.
+
+NOTE: As of version `5.0.0` of this plugin, a number of previously deprecated settings related to SSL have been removed.
+Please check out <<plugins-{type}s-{plugin}-obsolete-options>> for details.
+
+NOTE: As of version `5.0.0` of this plugin, a number of previously deprecated settings related to SSL have been removed.
+Please check out <<plugins-{type}s-{plugin}-obsolete-options>> for details.
 
 [cols="<,<,<",options="header",]
 |=======================================================================
@@ -478,6 +484,8 @@ Enable SSL/TLS secured communication to Elasticsearch cluster.
 Leaving this unspecified will use whatever scheme is specified in the URLs listed in <<plugins-{type}s-{plugin}-hosts>> or extracted from the <<plugins-{type}s-{plugin}-cloud_id>>.
 If no explicit protocol is specified plain HTTP will be used.
 
+When not explicitly set, SSL will be automatically enabled if any of the specified hosts use HTTPS.
+
 [id="plugins-{type}s-{plugin}-ssl_key"]
 ===== `ssl_key`
   * Value type is <<path,path>>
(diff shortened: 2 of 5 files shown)


donoghuc:
Got a failure on 7.x https://app.travis-ci.com/github/logstash-plugins/logstash-input-elasticsearch/jobs/628037127 that looked unrelated and suspiciously like a race condition. I reloaded that cell.

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -404,10 +393,9 @@ def validate_authentication
   def setup_client_ssl
     ssl_options = {}
     ssl_options[:ssl] = true if @ssl_enabled
+    ssl_options[:trust_strategy] = trust_strategy_for_ca_trusted_fingerprint
```
donoghuc, line 396: oops, i should only set this if `@ssl_enabled`. I'll move that back to where it was before.

robbavey (changes requested):
Partial review while we finalize the exact wording in docs, but let's get started with marking old settings as `obsolete` before removal

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -198,23 +198,12 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
   # Set the address of a forward HTTP proxy.
   config :proxy, :validate => :uri_or_empty
 
-  # SSL
```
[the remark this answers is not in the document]
donoghuc: ah, I totally missed that. Will get this fixed up!

donoghuc:
TODO:

- [ ] Review https://github.com/logstash-plugins/logstash-input-http/pull/182/files and make sure we have consistency where possible
- [ ] Grep through the repo for the old settings (including docs, tests and comments) to find references to remove/update

spec/inputs/elasticsearch_spec.rb
```diff
@@ -52,6 +52,19 @@
       end
     end
 
+    describe 'handling obsolete settings' do
```
donoghuc, line 55: Inspired by tests in https://github.com/logstash-plugins/logstash-input-http/pull/182 credit @robbavey

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -264,6 +253,11 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
   # If set, the _source of each hit will be added nested under the target instead of at the top-level
   config :target, :validate => :field_reference
 
+  # Obsolete Settings
+  config :ssl, :validate => :boolean, :default => false, :obsolete => "Set 'ssl_enabled' instead."
```
robbavey, line 257: Thinking about this, we can probably skip the `:validate` and `:default` parts, and stick with `:obsolete` to avoid users fixing up settings, only to find they were obsolete anyway. WDYT?

(My only caveat is to keep `:password` and `:uri` (not applicable here), as they are considered secret, and are [obfuscated](https://github.com/elastic/logstash/blob/a94659cf82d7e62c5a5f550df3e4e5bd0c6cf3e1/logstash-core/lib/logstash/config/mixin.rb#L608-L615) on discovery)

donoghuc: Nice, yeah i got tripped up on this writing tests. I was surprised to find that `:validate` would trigger when the option is `:obsolete`. I may be misunderstanding the test scenario but it looks to me like as a user I may be frustrated with a case whereby i set the wrong thing, then get a validation error that its the wrong form, only to be rewarded with a "this is no longer a valid option" once I satisfy the input validation. Not a huge deal but certainly a surprise if that is the case.

docs/index.asciidoc
```diff
@@ -608,12 +610,12 @@ option when authenticating to the Elasticsearch server. If set to an
 empty string authentication will be disabled.
 
 
-[id="plugins-{type}s-{plugin}-deprecated-options"]
-==== Elasticsearch Input deprecated configuration options
+[id="plugins-{type}s-{plugin}-removed-options"]
```
robbavey, line 613: For the docs update, let's follow the example in https://github.com/logstash-plugins/logstash-output-http/commit/2fae93e26b103cf9e7c0ec0d257be903070805b9#diff-cae5619b3d18ec99c5ccd0a9f6de0c6d3f53343c64692444551a7d29da6863e7

ie, remove all inline references to deprecated settings, and add a dedicated section at the bottom, above "common options", with a link to that section just before the config table.

cc @karenzone

donoghuc:
Rebased and updated documentation to match https://github.com/logstash-plugins/logstash-output-http/pull/147

donoghuc:
Corresponding docs PR https://github.com/elastic/logstash/pull/16744 in LS

robbavey (changes requested):
Copy/paste error and a comment nit

docs/index.asciidoc
```diff
@@ -608,47 +613,21 @@ option when authenticating to the Elasticsearch server. If set to an
 empty string authentication will be disabled.
 
 
-[id="plugins-{type}s-{plugin}-deprecated-options"]
-==== Elasticsearch Input deprecated configuration options
+[id="plugins-{type}s-{plugin}-obsolete-options"]
+==== HTTP Output Obsolete Configuration Options
```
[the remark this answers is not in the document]
donoghuc: Yep, fixed in amended commit. Good catch.

robbavey:
Should be good to go after a rebase

donoghuc:
Pushed the rebase. I think outstanding work after merge is:

[] Publish 5.0.0
[] PIn to 4 series in LS 8.x [example](https://github.com/elastic/logstash/commit/1d82e377ffcd63c7fad855d1f0cb6ab77398dfde)
[] Ensure [changes are documented](https://github.com/elastic/logstash/pull/16744) in LS

robbavey (approved):
LGTM - over to you @karenzone to check out doc changes

docs/index.asciidoc
```diff
@@ -98,6 +98,9 @@ TIP: Set the `target` option to avoid potential schema conflicts.
 
 This plugin supports the following configuration options plus the <<plugins-{type}s-{plugin}-common-options>> and the <<plugins-{type}s-{plugin}-deprecated-options>> described later.
```
karenzone, line 99:
```suggestion
This plugin supports these configuration options plus the <<plugins-{type}s-{plugin}-common-options>>  described later.
```

karenzone: This suggestion:
* Removes antiquated language ("the following..." )
* Removes inaccurate info suggesting that we still support deprecated/obsolete settings
* Removes invalid reference so that we won't break the docs builds

karenzone (changes requested):
Thanks for your work on this, @donoghuc. I left suggestion inline.

donoghuc:
Thanks @karenzone I committed your suggestion.

karenzone:
@donoghuc, one more thing I noticed after I got the docs building successfully. 
Looks like paragraph starting at line 633 was orphaned with removals.  

> Option to validate the server's certificate. Disabling this severely compromises security.
When certificate validation is disabled, this plugin implicitly trusts the machine
resolved at the given address without validating its proof-of-identity.
In this scenario, the plugin can transmit credentials to or process data from an untrustworthy
man-in-the-middle or other compromised infrastructure.
More information on the importance of certificate verification:
**https://www.cs.utexas.edu/~shmat/shmat_ccs12.pdf**.

Should it be deleted as part of `ssl_certificate_verification`?

donoghuc:
You are right! I just deleted that and squashed all the commits to clean up before merging :)

docs/index.asciidoc
```diff
@@ -96,7 +96,13 @@ TIP: Set the `target` option to avoid potential schema conflicts.
...
 ==== Elasticsearch Input configuration options
 
-This plugin supports the following configuration options plus the <<plugins-{type}s-{plugin}-common-options>> and the <<plugins-{type}s-{plugin}-deprecated-options>> described later.
+This plugin supports these configuration options plus the <<plugins-{type}s-{plugin}-common-options>> described later.
+
+NOTE: As of version `5.0.0` of this plugin, a number of previously deprecated settings related to SSL have been removed.
+Please check out <<plugins-{type}s-{plugin}-obsolete-options>> for details.
+
+NOTE: As of version `5.0.0` of this plugin, a number of previously deprecated settings related to SSL have been removed.
+Please check out <<plugins-{type}s-{plugin}-obsolete-options>> for details.
```
karenzone, line 105:
```suggestion
```

karenzone: Duplicate note

karenzone (approved):
Please delete duplicate note. Otherwise, LGTM!

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
-## 4.20.5
```
donoghuc, line 1: Need to make sure this is not deleted.

PR #215, opened 2024-11-20
Update .travis.yml to properly separate testing stages

jsvd:
If we don't include the `env` matrix row in the stage definition entry, then the previous entry is put under the next stage.

We can see this where a non integration test shows up in the integration test stage:

Build: https://app.travis-ci.com/github/logstash-plugins/logstash-input-elasticsearch/builds/273186251

Image:
Screenshot 2024-11-20 at 18 24 41

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,12 +7,12 @@ before_install:
 jobs:
   include:
     - stage: "Integration Tests"
-    - env: INTEGRATION=true SNAPSHOT=true LOG_LEVEL=info ELASTIC_STACK_VERSION=7.current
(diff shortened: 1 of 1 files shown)


jsvd:
The issue doesn't happen any more with this change:

Build: https://app.travis-ci.com/github/logstash-plugins/logstash-input-elasticsearch/builds/273233211

Image:
Screenshot 2024-11-20 at 18 27 43

donoghuc (approved):
Oops! Yes you are right. The docs even have this explicitly in an example https://docs.travis-ci.com/user/build-stages/#name-build-stages 

This is strange to my brain's internal yaml parser 😅

PR #223, opened 2025-03-04
Adds elastic-transport client support

mashhurs:
Adds elastic-transport client support where opens LS-core a way to start using newer ES ruby client.

FYI: failed [Test plugin docs](https://github.com/logstash-plugins/logstash-input-elasticsearch/actions/runs/13647730753?pr=223) CI is not related to this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 5.0.2
+  - Adds elastic-transport support [#223](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/223)
+
 ## 5.0.1
   - Fix: prevent plugin crash when hits contain illegal structure [#218](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/218)
     - When a hit cannot be converted to an event, the input now emits an event tagged with `_elasticsearch_input_failure` with an `[event][original]` containing a JSON-encoded string representation of the entire hit.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## 5.0.2
-  - Add elastic-transport client support [#223](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/223)
+  - Add elastic-transport client support used in elasticsearch-ruby 8.x [#223](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/223)
 
 ## 5.0.1
   - Fix: prevent plugin crash when hits contain illegal structure [#218](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/218)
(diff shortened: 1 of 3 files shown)


lib/logstash/inputs/elasticsearch.rb
```diff
@@ -642,6 +640,16 @@ def setup_query_executor
                       end
   end
 
+  def get_transport_client_class
+    require "elasticsearch/transport/transport/http/manticore"
+    require_relative "elasticsearch/patches/_elasticsearch_transport_http_manticore"
```
mashhurs, line 645: review note: 
- `_elasticsearch_transport_http_manticore` sets 
set user-agent and accept-encoding headers for the es-transport-client versions between 7.2 and 7.16. It seems to me we can remove this but I have kept so far as there are active 7.x users;
- `_elasticsearch_transport_connections_selector` has a round-robin load balancer logic for the es-client versions <7.2.0

I have kept them for now but it looks to me we can remove.

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -642,6 +640,16 @@ def setup_query_executor
                       end
   end
 
+  def get_transport_client_class
```
jsvd, line 643: let's add a comment here explaining what's happening

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 5.0.2
+  - Add elastic-transport client support [#223](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/223)
```
jsvd, line 2:
```suggestion
  - Add elastic-transport client support used in elasticsearch-ruby 8.x [#223](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/223)
```

jsvd (commented):
This is a good opportunity to be more explicit about the versions of es ruby this plugin supports, so instead of:

```
  s.add_runtime_dependency 'elasticsearch', '>= 7.17.9'
```

we could do

```
  s.add_runtime_dependency 'elasticsearch', '>= 7.17.9', '< 9'
```

as we haven't tested with 9 at all.

lib/logstash/inputs/elasticsearch.rb
mashhurs: Just added.

mashhurs:
> This is a good opportunity to be more explicit about the versions of es ruby this plugin supports, so instead of:
> 
> ```
>   s.add_runtime_dependency 'elasticsearch', '>= 7.17.9'
> ```
> 
> we could do
> 
> ```
>   s.add_runtime_dependency 'elasticsearch', '>= 7.17.9', '< 9'
> ```
> 
> as we haven't tested with 9 at all.

Yes, this is good point!
Updated. Thank you.

jsvd (approved):
LGTM

PR #226, opened 2025-03-17
Fix the unit test failures caused by #223

mashhurs:
Fixes the unit test failures caused by #223.
The failures became visible once LS-core upgraded elasticsearch-ruby client to 8.x.

### DIFF
--- a/spec/inputs/elasticsearch_spec.rb
+++ b/spec/inputs/elasticsearch_spec.rb
@@ -21,6 +21,13 @@ describe LogStash::Inputs::Elasticsearch, :ecs_compatibility_support do
   let(:es_version) { "7.5.0" }
   let(:cluster_info) { {"version" => {"number" => es_version, "build_flavor" => build_flavor}, "tagline" => "You Know, for Search"} }
 
+  def elastic_ruby_v8_client_available?
+    Elasticsearch::Transport
+    false
+  rescue NameError # NameError: uninitialized constant Elasticsearch::Transport if Elastic Ruby client is not available
+    true
+  end
+
   before(:each) do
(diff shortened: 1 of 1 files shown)


spec/inputs/elasticsearch_spec.rb
```diff
@@ -92,9 +92,11 @@
...
         before do
           allow(Elasticsearch::Client).to receive(:new).and_return(es_client)
-          allow(es_client).to receive(:info).and_raise(
-            Elasticsearch::Transport::Transport::Errors::BadRequest.new
-          )
+          begin
+            allow(es_client).to receive(:info).and_raise(Elasticsearch::Transport::Transport::Errors::BadRequest.new)
+          rescue NameError # NameError: uninitialized constant Elasticsearch::Transport
+            allow(es_client).to receive(:info).and_raise(Elastic::Transport::Transport::Errors::BadRequest.new)
+          end
```
andsel, line 99: I think that being more explicit on the intention would help to understand better:
```ruby

def elastic_ruby_client8_available?
  Elasticsearch::Transport
  return false
rescue NameError # NameError: uninitialized constant Elasticsearch::Transport if Elastic Ruby client is not available
  return true
end


before do
  ...
  unless elastic_ruby_client8_available?
    allow(es_client).to receive(:info).and_raise(Elasticsearch::Transport::Transport::Errors::BadRequest.new)
  else
    allow(es_client).to receive(:info).and_raise(Elastic::Transport::Transport::Errors::BadRequest.new)
  end
  ...
```

I know it's more verbose but maybe more explicit for the reader.

andsel (commented):
Left a comment to improve readability

andsel (approved):
LGTM

PR #234, opened 2025-04-07
Introduce cursor tracking akin to jdbc input (#205)

jsvd:
Provide field value tracking, persisted to disk on each search_after page.

Adds `:last_value` and `:present` placeholders, allowing the plugin to inject the cursor value and now-30 seconds, respectively, in the query string.

Useful to track new data being written to an index or series of indices.

Works best with nano second precision timestamps added by Elasticsearch's Ingest Pipelines.

backport #205 to 4.x


kaisecheng (approved):
LGTM

PR #235, opened 2025-06-06
ES|QL support

mashhurs:
Cherry picking https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/233 and updating branch/version specific changelog and version.

ES|QL support:

- introduces query_type params, accepts dsl or esql option.
- adds ES|QL executor to execute ESQL query and parse/map response to event validations
- make sure LS (8.17.4+) supports ES|QL (new elasticsearch-ruby client)
- make sure connected ES is greater than 8.11+
- query isn't empty or meaningful that starts with command syntax
- if query_type is esql, make sure we accept meaningful inputs and do not allow response_type, index, etc.. DSL related params
- informing if query isn't using METADATA which adds _id, _version to the response entries
- informing ineffective params such as size, search_api, target if users configure ES|QL results field names in a dotted format. The plugin reproduces nested (example {a.b.c: 'val'} => {'a':{'b':{'c':'val'}}})

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
-## 5.2.0
-  - ES|QL support [#233](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/233)
+## 4.23.0
+  - ES|QL support [#235](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/235)
 
 ## 4.22.0
   - Add "cursor"-like index tracking [#205](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/205)


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 5.2.0
+  - ES|QL support [#233](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/233)
+
```
mashhurs, line 3:
```suggestion
## 4.23.0
  - ES|QL support [#235](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/235)

```

PR #237, opened 2025-09-17
Support both, encoded and non encoded api-key formats on plugin configuration

### LINKED ISSUE #236: Support opaque `api_key`
The Elasticsearch input accepts an api_key parameter, which is expected to be in the name:key format. This value is base64-encoded into an opaque value before being used in an Authorization: ApiKey #{opaque} HTTP header.

To reduce friction, we should accept the opaque value for api_key anywhere that we accept the name:key format api_key.

We can do so by assuming that the value is opaque unless it includes a colon (:) character.

alexcams:
Avoid double-encoding opaque Elasticsearch API keys.

Previously, API keys that were already in encoded format were incorrectly re-encoded, causing connection failures. This update ensures that only raw keys are encoded, supporting both opaque and id:api-key formats without issue.

Closes #236 

### DIFF
--- a/lib/logstash/inputs/elasticsearch.rb
+++ b/lib/logstash/inputs/elasticsearch.rb
@@ -563,10 +563,18 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
   end
 
   def setup_api_key(api_key)
-    return {} unless (api_key && api_key.value)
+    return {} unless (api_key&.value)
 
-    token = ::Base64.strict_encode64(api_key.value)
+    token = is_base64?(api_key.value) ?  api_key.value : Base64.strict_encode64(api_key.value)
     { 'Authorization' => "ApiKey #{token}" }
+  end
+
+  def is_base64?(string)
+    begin
+      string == Base64.strict_encode64(Base64.strict_decode64(string))
+    rescue ArgumentError
+      false
+    end
   end
 
   def prepare_user_agent
--- a/spec/inputs/elasticsearch_spec.rb
+++ b/spec/inputs/elasticsearch_spec.rb
@@ -823,14 +823,26 @@ describe LogStash::Inputs::Elasticsearch, :ecs_compatibility_support do
       end
 
       context "with ssl" do
-        let(:config) { super().merge({ 'api_key' => LogStash::Util::Password.new('foo:bar'), "ssl_enabled" => true }) }
+        let(:config) { super().merge("ssl_enabled" => true) }
+        encoded_api_key = Base64.strict_encode64('foo:bar')
 
-        it "should set authorization" do
-          plugin.register
-          client = plugin.send(:client)
-          auth_header = extract_transport(client).options[:transport_options][:headers]['Authorization']
+        scenarios = {
+          'with non-encoded api-key' => LogStash::Util::Password.new('foo:bar'),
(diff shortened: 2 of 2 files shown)


lib/logstash/inputs/elasticsearch.rb
```diff
@@ -563,12 +563,20 @@ def setup_basic_auth(user, password)
...
   end
 
+  def is_base64?(string)
+    begin
+      string == Base64.strict_encode64(Base64.strict_decode64(string))
+    rescue ArgumentError
+      false
+    end
+  end
+
```
andsel, line 579: https://github.com/rubocop/ruby-style-guide?tab=readme-ov-file#implicit-begin

```suggestion
  def is_base64?(string)
    string == Base64.strict_encode64(Base64.strict_decode64(string))
  rescue ArgumentError
    false
  end

```

spec/inputs/elasticsearch_spec.rb
```diff
@@ -823,14 +823,26 @@ def synchronize_method!(object, method_name)
...
-        it "should set authorization" do
-          plugin.register
-          client = plugin.send(:client)
-          auth_header = extract_transport(client).options[:transport_options][:headers]['Authorization']
+        scenarios = {
+          'with non-encoded api-key' => LogStash::Util::Password.new('foo:bar'),
+          'with encoded api-key' => LogStash::Util::Password.new(encoded_api_key)
+        }
+
+        scenarios.each do |description, api_key_value|
```
andsel, line 834: Instead of dynamic context, use the explicit ones, and to reuse test code use `shared_example`

spec/inputs/elasticsearch_spec.rb
```diff
@@ -823,14 +823,26 @@ def synchronize_method!(object, method_name)
...
+          context description do
+            let(:config) { super().merge('api_key' => api_key_value) }
 
-          expect( auth_header ).to eql "ApiKey #{Base64.strict_encode64('foo:bar')}"
+            it "should set authorization" do
+              plugin.register
+              client = plugin.send(:client)
```
andsel, line 840: Could it use the `client` method directly without resort to the dynamic `send` ? Any reason for it?

andsel (changes requested):
Left some notes about code style.

spec/inputs/elasticsearch_spec.rb
alexcams: Done!

lib/logstash/inputs/elasticsearch.rb
alexcams: Done!

spec/inputs/elasticsearch_spec.rb
alexcams: it is a private method:
https://github.com/logstash-plugins/logstash-input-elasticsearch/blob/main/lib/logstash/inputs/elasticsearch.rb#L635

andsel: Correct, it's private because of the modifier https://github.com/logstash-plugins/logstash-input-elasticsearch/blob/973450396f0072954c1c29de68ea64b9a6e0e864/lib/logstash/inputs/elasticsearch.rb#L399.

So `send` is the way to bypass visibility rules.

PR #238, opened 2025-09-18
Remove non available stack release versions from integration CI tests

alexcams:
Plugin CI is broken as we can see in [this PR](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/237/checks?check_run_id=50577134130). The reason is that the [plugin CI file](https://github.com/logstash-plugins/logstash-input-elasticsearch/blob/main/.travis.yml) includes stack release versions that are not available (`8.next and 8.future`). This is referenced at: https://github.com/logstash-plugins/.ci/blob/1.x/logstash-versions.yml.

First commit removes those unavailable releases from integration tests jobs definition to fix the CI and adds 9.x versions.


spec/inputs/paginated_search_spec.rb
```diff
@@ -1,5 +1,6 @@
 require "logstash/devutils/rspec/spec_helper"
 require "logstash/inputs/elasticsearch/paginated_search"
+require "manticore"
```
mashhurs, line 3: Hm., I don't see the manticore usage in this file or I can't find ([here](https://app.travis-ci.com/github/logstash-plugins/logstash-input-elasticsearch/pull_requests?serverType=git)) CIs failing due to related to this change.

mashhurs (commented):
Matrix LGTM!
One question I have raised if manticore dependency is required in the spec file.

spec/inputs/paginated_search_spec.rb
alexcams: Removed it! :)

PR #247, opened 2026-04-02
Remove rufus cheduler.

mashhurs:
Removes rufus-scheduler development dependency, new version will be consumed through logstash-mixin-scheduler, comatible with jruby 10, extra details - https://github.com/elastic/logstash/pull/14260


andsel (approved):
LGTM

PR #251, opened 2026-05-01
[Doc] search_after authorization for aliases

### LINKED ISSUE #250: Possible authz bug: Logstash search_after/PIT fails with API key while scroll succeeds
We may have found a bug or authz edge case involving Logstash 9.x, Elasticsearch 8.19.10, and API key auth when the Elasticsearch input auto-selects search_after.

Observed behavior

Logstash 9.x auto-resolves search_api => search_after for ES clusters >= 7.10
Early startup log shows:
search_api => auto resolved to search_after {:elasticsearch=>"8.19.10"}

With this mode, the input fails as if the API key does not have read privilege on the target index
The same API key works when we explicitly set:
search_api => "scroll"

Why this looks suspicious

The API key role descriptor grants read on the target index
scroll works with the same key and same index
Failure appears specific to the search_after / PIT path
Working theory

search_after uses PIT
In that flow, Logstash creates a PIT and then issues paginated searches via:
POST /_search

with the PIT id in the request body, rather than:
POST /<index>/_search

It appears authz is being resolved differently on that PIT-based path, and the API key is treated as if it lacks index read, even though the role descriptor grants it
Why scroll works

Forcing search_api => "scroll" reverts to requests like:
POST /<index>/_search?scroll=10m

Because the index is explicit in the URL, authz resolves cleanly and the backfill succeeds
Impact

Logstash 9.x can fail unexpectedly against ES 8.19.10 with API key auth when search_api auto-selects search_after
Users may interpret this as a bad API key or missing privileges, even though scroll proves the same key works
Current workaround is to force:
search_api => "scroll"

Ask

Can the team confirm whether this is a known issue in:
Elasticsearch authz for PIT + API key
or the Logstash Elasticsearch input plugin’s search_after behavior

If not known, can this be triaged as a potential bug/regression?

We noticed this behavior while working with a customer, I don't have additional logs on this.

kaisecheng:
Adds a NOTE under `search_api` explaining that when using `search_after`, read permissions on aliases alone are not enough. Permissions must be granted on the underlying indices or data streams.

Bumps version to 5.2.2.

Closes #250

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## 5.2.2
-  - [DOC] Note that `search_after` requires permissions on underlying indices/data streams, not aliases [#TBD](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/TBD)
+  - [DOC] Note that `search_after` requires permissions on underlying indices/data streams, not aliases [#251](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/TBD)
 
 ## 5.2.1
   - Added support for encoded and non encoded api-key formats on plugin configuration [#237](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/237)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 5.2.2
+  - [DOC] Note that `search_after` requires permissions on underlying indices/data streams, not aliases [#TBD](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/TBD)
+
 ## 5.2.1
   - Added support for encoded and non encoded api-key formats on plugin configuration [#237](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/237)
 
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -120,10 +120,10 @@ Examples include:
 
 The Elasticsearch input plugin provides the <<plugins-{type}s-{plugin}-tracking_field>> and <<plugins-{type}s-{plugin}-tracking_field_seed>> options.
 When <<plugins-{type}s-{plugin}-tracking_field>> is set, the plugin records the value of that field for the last document retrieved in a run into
-a file. 
+a file.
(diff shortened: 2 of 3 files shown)


docs/index.asciidoc
```diff
@@ -706,6 +706,8 @@ The query requires at least one `sort` field, as described in the <<plugins-{typ
 
 `scroll` uses {ref}/paginate-search-results.html#scroll-search-results[scroll] API to search, which is no longer recommended.
 
+NOTE: `search_after` needs read permissions on the underlying indices or data streams. Permissions on aliases alone are not enough.
```
robbavey: How about something like:

```suggestion
NOTE: When using `search_after`, including using `auto` on  Elasticsearch 8+, permissions to read data aliases alone will not work: permissions are also needed on the underlying indices or data streams.
```

robbavey (changes requested):

robbavey (approved):
lgtm

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 5.2.2
+  - [DOC] Note that `search_after` requires permissions on underlying indices/data streams, not aliases [#251](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/TBD)
```
mashhurs, line 2: Oops! we missed to update PR no. Good news is we don't have it referenced in our main elastic-docs but elastic-docs references to GH changelog and changelog has TBD link - https://github.com/logstash-plugins/logstash-input-elasticsearch/blob/v5.2.2/CHANGELOG.md
So, it seems will be okay if we update 5.2.2 tag's PR no.

Another thing, Kaise seems published the plugin -  https://rubygems.org/gems/logstash-input-elasticsearch/versions/5.2.2 but I still don't see the version in the VPR list - 
https://www.elastic.co/docs/reference/logstash/versioned-plugins/input-elasticsearch-index

- do we need to fix the PR no and release the plugin again or can we update just changelog of tags?
- how VPR docs build work now? - I feel out of docs-tools sync :( but in the past in 6-hours we had the fresh pages.

kaisecheng: Updated the correct link. Hopefully, VPR can get it

PR #252, opened 2026-05-05
Allow upgrading elasticsearch ruby client and remove its 7.x patch logics.

### LINKED ISSUE #248: Upgrade `elasticsearch` ruby client and remove 7.x logics if exist
### Description
Upgrade elasticsearch client in `4.x` and `main` active branches 
- `main`: https://github.com/logstash-plugins/logstash-input-elasticsearch/blob/d2ef49879799afd990cc917f2505112146160646/logstash-input-elasticsearch.gemspec#L29
- `4.x` - https://github.com/logstash-plugins/logstash-input-elasticsearch/blob/b89667940da18257bc37929cc6dbaa026930c635/logstash-input-elasticsearch.gemspec#L29
- remove/adapt patch logics if need - https://github.com/logstash-plugins/logstash-input-elasticsearch/tree/main/lib/logstash/inputs/elasticsearch/patches

mashhurs:
This PR drops a support for Logstash 7.x although plugin still can be installed.
- removes elasticsearch 7.x specific patch logics
- allow upgrading elasticsearch ruby client up to 9.x versions


mashhurs:
/pr_prepare release: minor
- Drop a support for Logstash 7.x by requiring `elasticsearch` gem >= 8. Logstash 8+ continues to work as before.

kaisecheng (commented):
Thanks for the cleanup! One concern on the version bump. I believe it should be at least a 'minor' because upgrading gem version is not a bug fix and this change will block 7.17 user to install. I tested in 7.17.25 and confirmed that user can still install es-input 5.2.2. 

I'd suggest `minor`/ `major` with a note in the changelog
```
  - Drop support for Logstash 7.x by requiring `elasticsearch` gem >= 8.
    Logstash 8+ continues to work as before.
```

mashhurs:
> Thanks for the cleanup! One concern on the version bump. I believe it should be at least a 'minor' because upgrading gem version is not a bug fix and this change will block 7.17 user to install. I tested in 7.17.25 and confirmed that user can still install es-input 5.2.2.
> 
> I'd suggest `minor`/ `major` with a note in the changelog
> 
> ```
>   - Drop support for Logstash 7.x by requiring `elasticsearch` gem >= 8.
>     Logstash 8+ continues to work as before.
> ```

Yep, just updated the [changelog/version-bump request](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/252#issuecomment-4435486568). As we discussed in the weekly, I will add release note when this change will be embedded in LS core and I have opened an issue if there anything more we need follow up - https://github.com/logstash-plugins/logstash-input-elasticsearch/issues/254

kaisecheng (approved):
LGTM

mashhurs:
/pr_prepare release:minor
- Drop a support for Logstash 7.x by requiring `elasticsearch` gem >= 8. Logstash 8+ continues to work as before.

PR #269, opened 2026-06-17
Fix serverless header conflict with elasticsearch-ruby v9

kaisecheng:
Although the elasticsearch-ruby gem fixed the header conflicts [issue](https://github.com/elastic/elasticsearch-ruby/issues/2983), the es-input serverless integration tests are [still failing](https://buildkite.com/elastic/logstash-serverless-integration-testing/builds/1151#019ed0b7-09c4-4c96-a2f9-3d5ff8dde3f2) with:

```
The request includes both the [Elastic-Api-Version] header and a
[compatible-with] parameter, but it is not valid to include both of
these in a request
```

### Root cause

This plugin creates the client twice. The first client checks if ES is serverless. If so, the plugin recreates the client by reusing the previous `transport_options` hash with an additional `Elastic-Api-Version` header.

However, the elasticsearch-ruby client mutated the headers by adding `compatible-with` during the first call, which leaks into the second.

### Fix

This PR builds client from fresh transport headers


kaisecheng:
Broken CI due to i18n gem new release https://github.com/ruby-i18n/i18n/issues/735

alexcams (approved):
LGTM

PR #274, opened 2026-06-23
Support Elastic Cloud API keys in the `api_key` option

### LINKED ISSUE #275: Support Elastic Cloud org level API keys
es-input plugin accepts 
- `<id>:<key>` pair API key (Logstash, beats format)
- base64 (obfuscated key:val) encoded string
Any other formats will be base64 encoded and used in the HTTP requests.
Elastic cloud org level APIs carry `essu_` prefix + `base64`  string which is neither `<id>:<key>` pair or `base64` format, so plugin base64 encodes this key and ES rejects.

Benefit of using Elastic cloud API is person independent but Hosted/Project API keys are tight to the person who creates the API keys.

Relates to https://github.com/elastic/logstash/issues/17360

### Expectation
When providing Elastic cloud org level APIs to `api_key` param, plugin needs to utilize as it is in its requests to connecting Elastic cloud.

arenard:
## Summary

Allows an Elastic Cloud API key (the `essu_`-prefixed keys, including the unified Serverless keys) to be used in the `api_key` option. These keys are not in the `id:api_key` form and cannot currently be used with this input.

## Background

The `api_key` value is turned into an `Authorization: ApiKey <value>` header. The existing logic base64-encodes any value that is not already valid standard base64. An `essu_` key is neither an `id:api_key` pair nor valid standard base64 (the `essu_` prefix contains `_`, which is outside the standard base64 alphabet), so it ends up being base64-encoded a second time and Elasticsearch rejects the resulting header.

Elasticsearch accepts these keys directly as `Authorization: ApiKey <key>` (Cloud API key authentication, available since Elasticsearch 9.1), so the gap is only on the Logstash side.

## Change

`setup_api_key` now recognises the `essu_` prefix and forwards such keys unchanged. Raw `id:api_key` pairs and their base64-encoded form keep their previous behaviour, so existing configurations are unaffected.

## Tests

Added a unit test that asserts an `essu_` key is sent verbatim, with no re-encoding. The existing raw and already-encoded api-key tests still pass.

## Docs

Updated `docs/index.asciidoc` and the inline `api_key` option documentation to list the three accepted forms.

## Related

The same fix is applied to the other two Elasticsearch plugins so the behaviour is consistent:
- logstash-plugins/logstash-output-elasticsearch#1274
- logstash-plugins/logstash-filter-elasticsearch#215


lib/logstash/inputs/elasticsearch.rb
```diff
@@ -542,10 +543,20 @@ def setup_basic_auth(user, password)
   def setup_api_key(api_key)
     return {} unless (api_key&.value)
 
-    token = base64?(api_key.value) ? api_key.value : Base64.strict_encode64(api_key.value)
+    # Elastic Cloud API keys (`essu_` prefixed) and already base64-encoded
+    # keys are passed verbatim; a raw `id:api_key` pair is encoded here.
+    token = (cloud_api_key?(api_key.value) || base64?(api_key.value)) ? api_key.value : Base64.strict_encode64(api_key.value)
     { 'Authorization' => "ApiKey #{token}" }
   end
```
alexcams: This logic is actually valid, however I think it could be easier to move this logic to a function and make it a bit easier to read. What do you think about something like: 
```suggestion
    token = resolve_api_key(api_key.value)
    { 'Authorization' => "ApiKey #{token}" }
  end
  
  def resolve_api_key(key_value)
    if base64?(key_value) || cloud_api_key?(key_value)
      key_value
    elsif key_value.include?(':')
      Base64.strict_encode64(key_value)
    else
      raise LogStash::ConfigurationError, "Invalid api_key format. Expected a base64-encoded key, an 'id:api_key' pair, or a Cloud API key (essu_ prefix)."
    end
  end
```
WDYT? 🙂

lib/logstash/inputs/elasticsearch.rb
```diff
@@ -542,10 +543,20 @@ def setup_basic_auth(user, password)
...
 
+  # Elastic Cloud API keys scoped for Elasticsearch (such as the unified
+  # Serverless keys) are opaque `essu_` prefixed tokens that Elasticsearch
+  # accepts verbatim in the `Authorization: ApiKey` header, with no base64
+  # encoding.
+  def cloud_api_key?(string)
+    string.start_with?("essu_")
+  end
```
alexcams, line 570: Cloud Api keys expected format is: `essu_<base64-encoded payload>`
We could make this more robust by checking if the content after the prefix is base64 valid to avoid errored api keys:
```suggestion
  def cloud_api_key?(string)
    return false unless string.start_with?("essu_")

    base64?(string.delete_prefix("essu_"))
  end
```

spec/inputs/elasticsearch_spec.rb
```diff
@@ -835,6 +835,18 @@ def synchronize_method!(object, method_name)
           it_behaves_like "a plugin that sets the ApiKey authorization header"
         end
 
+        context "with an Elastic Cloud API key (essu_ prefix)" do
+          let(:api_key_value) { "essu_VFZGblZreFhTekJ4ZDB4M2NHUnZRMEU2YzNWd1pYSnpaV055WlhRPQ==AAAAAAAA" }
```
alexcams, line 841: This example api key would fail with the implementation suggested abobe

docs/index.asciidoc
```diff
@@ -404,10 +404,12 @@ input plugins.
 
 Authenticate using Elasticsearch API key. Note that this option also requires enabling the <<plugins-{type}s-{plugin}-ssl_enabled>> option.
 
-Format is `id:api_key` where `id` and `api_key` are as returned by the
+The format is `id:api_key`, where `id` and `api_key` are as returned by the
 Elasticsearch
 {ref}/security-api-create-api-key.html[Create
-API key API].
+API key API]. The base64-encoded form of that pair is also accepted, as is an
```
alexcams, line 410: Thank you for the base64 clarification 🙂

alexcams (changes requested):
Hi @arenard! Thank you so much for all this contributions, very useful, with meaningful comments 🙂
I've suggested a couple of changes to make this more robust and help catch errors easily in the future. Let me know your thoughts about and if you agree, we'd appreciate if it's moved to the other 2 PRs.

spec/inputs/elasticsearch_spec.rb
```diff
@@ -835,6 +835,18 @@ def synchronize_method!(object, method_name)
...
+
+          it "sets the Authorization header verbatim without re-encoding" do
+            plugin.register
+            client = plugin.send(:client)
+            auth_header = client.transport.options[:transport_options][:headers]['Authorization']
+
+            expect(auth_header).to eql("ApiKey #{api_key_value}")
+          end
+        end
```
alexcams, line 850: I'd also add a case to test api keys starting with `essu_` and then invalid base64 encoding.

lib/logstash/inputs/elasticsearch.rb
arenard: Agreed, this reads much better and the explicit error surfaces a bad key at config time instead of a silent 401 later.

arenard: @alexcams on comment line 555 `elsif key_value.include?(':')` instead of checking if key contains a `':'`, maybe we could improve the test to ensure it contains a key validating the form `id:api_key`. 

I would suggest something like this to check there is only one `':'` in the key : 
`elsif key_value.count(':') == 1`
or checking the full format like : 
`elsif key_value.match?(/\A[^:]+:[^:]+\z/)`

WDYT? 🙂

lib/logstash/inputs/elasticsearch.rb
arenard: @alexcams Aligned on the path to validate early. 
I looked into it before touching `cloud_api_key?` and I'd lean toward keeping it as a prefix check with non-empty suffix only.
Are we sure the format is guaranted to be: `essu_<base64-encoded payload>` ? I could not find any public spec about it.

The reference UIAM client in Kibana only checks the prefix, not the base64 of the payload: [link](https://github.com/elastic/kibana/blob/8c43c76a8b76eedc622307c544466f9024c4251c/src/core/packages/security/server/src/uiam/utils.ts#L20-L24).

Looking at the Cloud side that introduced the `essu_`, prefix+payload format is lenient, if the prefix is missing or the payload is invalid it falls back to authenticating with the key as-is. 
It looks like the payload after `essu_` isn't guaranteed to be canonical standard base64. Requiring a `strict_encode64(strict_decode64(...))` round-trip would reject those, and combined with the new error branch would turn a potential valid key into a ConfigurationError.

WDYT? 🙂

arenard: Should we change the name of the function for something like : 
`uiam_credential?` instead of `cloud_api_key?` ?

lib/logstash/inputs/elasticsearch.rb
alexcams: I agree, both are better than the previous implementation. I don't have strong preference for any.

lib/logstash/inputs/elasticsearch.rb
alexcams: > Are we sure the format is: essu_<base64-encoded payload>?

Well, it is not documented as it, I just created 5 and all of them had that format, but continues being a guessing. I'd also follow the same implementation as kibana team, and same for the name. Thanks for clarification 🙂

alexcams (approved):

PR #277, opened 2026-07-14
[Backport] Remove Rufus scheduler

alexcams:
Removes rufus-scheduler development dependency, new version will be consumed through logstash-mixin-scheduler, comatible with jruby 10, extra details - https://github.com/elastic/logstash/pull/14260

Reference: https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/247

### DIFF
--- a/lib/logstash/inputs/elasticsearch.rb
+++ b/lib/logstash/inputs/elasticsearch.rb
@@ -312,8 +312,6 @@ class LogStash::Inputs::Elasticsearch < LogStash::Inputs::Base
   end
 
   def register
-    require "rufus/scheduler"
-
     @pipeline_id = execution_context&.pipeline_id || 'main'
 
     fill_hosts_from_cloud_id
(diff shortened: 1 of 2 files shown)


lib/logstash/inputs/elasticsearch.rb
```diff
@@ -312,8 +312,6 @@ def initialize(params={})
   end
 
   def register
-    require "rufus/scheduler"
```
mashhurs, line 315: I don't think we need to remove this usage. Only dev dependency should be removed.

mashhurs (changes requested):

lib/logstash/inputs/elasticsearch.rb
alexcams: I'm not really sure that we need that. The scheduler we use comes from https://github.com/alexcams/logstash-input-elasticsearch/blob/main/lib/logstash/inputs/elasticsearch.rb#L11

I think that require was residual code from https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/177/changes

mashhurs: oh makes sense. Can we also remove in `main` then? It looks like I missed in https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/247

mashhurs (approved):

lib/logstash/inputs/elasticsearch.rb
alexcams: Sure!

